PR #2, opened 2025-07-02
[18.0][MIG] shopfloor - round 1

sebalix:
Starting from not installable module #5 .

Depends on:
- https://github.com/OCA/stock-logistics-workflow/pull/1971
- https://github.com/OCA/stock-logistics-workflow/pull/2031

## TODO

Status of the migration:

- [x] core functions (actions, helpers)
- [x] handling of qty done
- [ ] single_pack_transfer (to be discarded in favor of shopfloor_single_product_transfer)
- [ ] cluster_picking
- [ ] checkout
- [ ] delivery
- [ ] location_content_transfer
- [ ] zone_picking


shopfloor/actions/stock.py
```diff
@@ -0,0 +1,252 @@
...
+        for move in origin_move.move_dest_ids:
+            if (
+                move.origin_returned_move_id
+                and move.origin_returned_move_id != origin_move
+            ):
+                continue
+            if move.state in ("partially_available", "assigned"):
+                quantity -= sum(move.move_line_ids.mapped("quantity_product_uom"))
+            elif move.state in ("done"):
+                quantity -= move.product_uom_qty
```
grindtildeath:
```suggestion
                quantity -= move.quantity
```
I guess the same field can be used for the condition above

simahawk:
As per internal discussion here's a PR w/ the module not installable to be able to focus here only on migration changes
https://github.com/OCA/stock-logistics-shopfloor/pull/5

simahawk:
Rebased after #5 and removed the test dep on https://github.com/OCA/stock-logistics-workflow/pull/1971 that was merged.

simahawk:
My proposal to ease devs for other modules too:
1. merge ASAP w/o scenario
2. list scenario as TODO list in the readme
3. work on each scenario one by one
The same applies to `shopfloor_mobile`.

NOTE: we discussed about splitting scenario out to single modules but it will make harder to fwd/bkp changes to/from v16.

shopfloor/actions/data.py
```diff
@@ -217,8 +217,9 @@ def move_lines(self, records, **kw):
     def _move_line_parser(self):
         return [
             "id",
-            "qty_done",
-            "reserved_uom_qty:quantity",
+            "picked",
```
simahawk: these changes that affect the response schema are not good.
We should rollback and
* keep `qty_done`, if necessary map it to another field or add a computed field
* drop `picked` and `quantity_picked`

CC @grindtildeath

grindtildeath: Let's check my changes there before having it merged here
https://github.com/camptocamp/stock-logistics-shopfloor/pull/3

shopfloor/actions/data.py
```diff
@@ -218,7 +218,8 @@ def _move_line_parser(self):
         return [
             "id",
             "qty_done",
-            "reserved_uom_qty:quantity",
+            # FIXME
```
simahawk: fixme what?

shopfloor/actions/data.py
```diff
@@ -227,7 +228,8 @@ def _move_line_parser(self):
                 "move_id:priority",
                 lambda rec, fname: rec.move_id.priority or "",
             ),
-            "progress",
+            # TODO: progress is not available on move lines anymore, what to do for UI?
```
simahawk, line 230: doesn't come from `stock_picking_progress`?

test-requirements.txt
```diff
@@ -0,0 +1 @@
+odoo-addon-stock_move_line_change_lot @ git+https://github.com/OCA/stock-logistics-workflow@refs/pull/1971/head#subdirectory=stock_move_line_change_lot
```
simahawk: this should be updated

simahawk (commented):
LGTM overall

shopfloor/actions/data.py
```diff
@@ -358,7 +360,7 @@ def _get_operation_progress(self, domain):
         operations_done = 0
         for line in lines:
             operations_done += line.qty_done
-            operations_to_do += line.reserved_uom_qty - line.qty_done
+            operations_to_do += line.quantity - line.qty_done
```
simahawk: shouldn't be `line.qty_picked`?

simahawk:
> There's no issue in this repo with the title 'Migration to version 18.0' and the milestone 18.0, so not possible to add the comment.

you are drunk bro!

simahawk:
Last thing to fix
```
2025-07-31 10:50:59,300 256 WARNING odoo odoo.addons.shopfloor.models.stock_move_line: `_split_pickings_from_source_location` is deprecated and replaced by `_extract_in_split_order` 
```

simahawk:
> My proposal to ease devs for other modules too:
> 
>     1. merge ASAP w/o scenario
> 
>     2. list scenario as TODO list in the readme
> 
>     3. work on each scenario one by one
>        The same applies to `shopfloor_mobile`.
> 
> 
> NOTE: we discussed about splitting scenario out to single modules but it will make harder to fwd/bkp changes to/from v16.

FTR I reported this strategy on the README of the module and on #13

shopfloor/__manifest__.py
```diff
@@ -6,7 +6,7 @@
 {
     "name": "Shopfloor",
     "summary": "manage warehouse operations with barcode scanners",
-    "version": "16.0.2.8.0",
+    "version": "18.0.1.0.0",
```
grindtildeath: Wondering if we shouldn't have a 18.0.0.x.y as we didn't _fully_ migrate yet :question:

grindtildeath (approved):
I only checked the last few commits and IMO it looks good for a first version, we can still have more fixes coming as we are migrating scenarios and other modules

shopfloor/__manifest__.py
simahawk: Kind of make sense :thinking: :+1:

simahawk: done 0.1.0

PR #3, opened 2025-07-11
[18.0] Create module stock_move_line_qty_picked

grindtildeath:
This module adds a ``qty_picked`` field on ``stock.move.line`` to allow
scanning different quantities without having to update the ``quantity``
field, what would modify existing reservations.

It is meant to restore the behaviour we had in previous versions with
``qty_done`` field, but using another name to avoid clashing with the
field defined in Odoo Enterprise's Stock barcode module.


stock_move_line_qty_picked/readme/DESCRIPTION.md
```diff
@@ -0,0 +1,3 @@
+This module adds a  `qty_picked` field on `stock.move.line` to allow scanning
+different quantities without having to update the `quantity` field, what would
+modify existing reservations.
```
simahawk, line 3: I'd mention that this is to restore the behavior prior to v18, no?

stock_move_line_qty_picked/models/stock_move_line.py
```diff
@@ -0,0 +1,33 @@
+# Copyright 2025 Camptocamp SA
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl)
+from odoo import fields, models
+from odoo.tools import float_compare
+
+
+class StockMoveLine(models.Model):
+    _inherit = "stock.move.line"
+
+    qty_picked = fields.Float()
```
simahawk, line 10: I'd mention that this name has been chosen also not to clash w/ EE.

simahawk (commented):
LG. This probably goes to `stock-logistics-workflow`?

stock_move_line_qty_picked/models/stock_move_line.py
```diff
@@ -0,0 +1,33 @@
...
+
+    def _pick_qty(self, qty):
+        values = {
+            "qty_picked": qty,
+            "picked": True,
+        }
+        total_demand = self.move_id.product_uom_qty
+        total_reserved = sum(self.move_id.move_line_ids.mapped("quantity"))
+        prec = self.env["decimal.precision"].precision_get("Product Unit of Measure")
```
TDu, line 19: Should use the precision on the uom of the product/move ?

stock_move_line_qty_picked/models/stock_move_line.py
```diff
@@ -0,0 +1,33 @@
...
+from odoo.tools import float_compare
+
+
+class StockMoveLine(models.Model):
+    _inherit = "stock.move.line"
+
+    qty_picked = fields.Float()
+
+    def _pick_qty(self, qty):
+        values = {
```
TDu, line 13:
```suggestion
        self.ensure_one()
        values = {
```

stock_move_line_qty_picked/models/stock_move.py
```diff
@@ -0,0 +1,52 @@
...
+            picked_qty = 0
+            for ml in self.move_line_ids:
+                picked_qty += ml.product_uom_id._compute_quantity(
+                    ml.qty_picked, self.product_uom, round=False
+                )
+            return picked_qty
+        return super()._get_picked_quantity()
+
+    @api.depends("move_line_ids.picked", "move_line_ids.qty_picked")
+    def _compute_quantity(self):
```
TDu, line 22: Why not a different field for computing the picked quantity ?

grindtildeath: We actually want the quantity field to reflect the picked_qty in the picking validation process because Odoo does some checks based on this field.

stock_move_line_qty_picked/models/stock_move_line.py
grindtildeath: AFAIU we cannot have a precision on the UOM that is "smaller" than the one defined on the decimal.precision record, so I guess it must be the same? 

Maybe that's a topic to research for SATK @gurneyalex :question:

grindtildeath:
Replaced by https://github.com/OCA/stock-logistics-workflow/pull/2031

PR #5, opened 2025-07-21
[18.0][MIG] shopfloor: port as-is no install

simahawk:


TDu (approved):

grindtildeath (approved):
Yes please :pray:

mmequignon:
SHouldn't we do that from @sebalix branch instead? He managed to have lots of things working already….
ping @TDu @grindtildeath @simahawk

grindtildeath:
@mmequignon we don't want to merge #2 yet, we only want to merge the non migrated module's code, so that the migration PR (ie #2) shows only the migration commits + diff, instead of the whole history

simahawk:
> SHouldn't we do that from @sebalix branch instead? He managed to have lots of things working already…. ping @TDu @grindtildeath @simahawk

Nice to see you were fully with us in the call :rofl: :stuck_out_tongue:

mmequignon:
> > SHouldn't we do that from @sebalix branch instead? He managed to have lots of things working already…. ping @TDu @grindtildeath @simahawk
> 
> Nice to see you were fully with us in the call 🤣 😛

Well yeah, but if we start working now against 18.0, we wont base our work on what Sebastien already did, and might end up fixing twice the same things.

PR #7, opened 2025-07-25
[18.0] [MIG] shopfloor_delivery_shipment_mobile

TDu:


simahawk (commented):
some comments can be removed, the linting should happen before the  migration commit

simahawk:
@TDu just to check: have you tested the demo mode?

simahawk (approved):

TDu:
> @TDu just to check: have you tested the demo mode?

I tested with real data, not with the demo mode.

simahawk:
I'll check that. Demo mode is very useful when developing and migrating.

PR #19, opened 2025-08-04
[18.0][MIG] shopfloor: ``single_pack_transfer`` scenario

grindtildeath:
Requires:
 - [ ] <s>https://github.com/odoo/odoo/pull/219995/</s>
 - [x] https://github.com/odoo/odoo/pull/222034

### DIFF
--- a/shopfloor/services/single_pack_transfer.py
+++ b/shopfloor/services/single_pack_transfer.py
@@ -118,7 +118,7 @@ class SinglePackTransfer(Component):
                     ("state", "in", ("assigned", "partially_available")),
                 ]
             )
-            if any(line.qty_done > 0 for line in other_move_lines) or (
+            if any(line.qty_picked > 0 for line in other_move_lines) or (
                 other_move_lines and not self.work.menu.allow_unreserve_other_moves
             ):
                 picking = fields.first(other_move_lines).picking_id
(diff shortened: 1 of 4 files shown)


shopfloor/tests/test_single_pack_transfer.py
```diff
@@ -804,20 +804,37 @@ def test_cancel_transfer_not_created_by_user(self):
...
+        self.assertRecordValues(self.picking, [{"state": "assigned"}])
+        for move in self.picking.move_ids:
+            self.assertRecordValues(move, [{"state": "assigned"}])
+        for pack_level in self.picking.package_level_ids:
+            self.assertFalse(pack_level.is_done)
+
         package_level = self._simulate_started(self.pack_a)
+
         self.menu.sudo().allow_move_create = True
         self.env.user = self.shopfloor_manager
```
sebalix, line 946: Maybe this wobbly line could be removed now that you are bootstrapping a dedicated service with shopfloor manager user?

sebalix (commented):
LG, one small comment.

And should we prepare a workaround in case Odoo doesn't fix the core issue?


EDIT: there is this line that should be fixed in `single_pack_transfer.py`:
```diff
-            if any(line.qty_done > 0 for line in other_move_lines) or (
+            if any(line.picked for line in other_move_lines) or (
```

grindtildeath:
As we discussed with @simahawk and @sebalix , the issue is following:

When we remove the `is_done` flag on `stock.package_level`, Odoo systematically deletes the related move line, even if previously we had a move line linked to the package level while `is_done` was False. Without my fix in Odoo, this would leave the stock move flagged as picked, and in turn we cannot assign the move to have a move line created again.

To check, when starting, instead of flagging the stock package level as `is_done=True`, shouldn't we mark the linked move lines as `picked=True` ? Then we wouldn't have to touch `is_done` and we can keep the existing move line in case we cancel the operation from shopfloor

henrybackman (approved):

shopfloor/tests/test_single_pack_transfer.py
sebalix: Is this line still needed?

sebalix (commented):
There are still occurrences of `qty_done` field to replace in `test_single_pack_transfer.py` file (see CI)

sebalix:
PR rebased and fixup squashed.

shopfloor/services/single_pack_transfer.py
```diff
@@ -286,18 +286,18 @@ def cancel(self, package_level_id):
...
+        if self.is_allow_move_create() and picking.create_uid == self.env.user:
             # Cancel the transfer when it has been created by the shopfloor user
-            moves.picking_id.action_cancel()
-        else:
-            # Not owned only unassign the user
-            stock = self._actions_for("stock")
-            stock.unmark_move_line_as_picked(moves.move_line_ids)
+            picking.action_cancel()
+        elif assigned_picking:
+            # FIXME: Requires https://github.com/odoo/odoo/pull/219995
```
TDu: Fixme can be removed

simahawk: dropped and rebased

PR #22, opened 2025-08-06
[18.0][MIG] shopfloor: `cluster_picking` scenario

mmequignon:


shopfloor/services/cluster_picking.py
```diff
@@ -865,7 +865,7 @@ def _data_for_unload_single(self, batch, package, confirmation=None):
     def _filter_for_unload(self, line):
         return (
             line.state in ("assigned", "partially_available")
-            and line.qty_done > 0
+            and line.qty_picked > 0
```
sebalix:
```suggestion
            and line.picked
```

shopfloor/services/cluster_picking.py
```diff
@@ -1035,7 +1042,7 @@ def _domain_stock_issue_unlink_lines(self, move_line):
             ("package_id", "=", package.id),
             ("lot_id", "=", lot.id),
             ("state", "not in", ("cancel", "done")),
-            ("qty_done", "=", 0),
+            ("qty_picked", "=", 0),
```
sebalix:
```suggestion
            ("picked", "=", False),
```

shopfloor/tests/test_cluster_picking_base.py
```diff
@@ -35,8 +35,8 @@ def _line_data(self, move_line, qty=None, package_dest=False, qty_done=None, **k
             data["package_dest"] = None
         if qty:
             data["quantity"] = qty
-        if qty_done:
-            data["qty_done"] = qty_done
+        if qty_picked:
+            data["qty_picked"] = qty_picked
```
sebalix: We should probably not change `qty_done` here, it's data we are sending to the client.

shopfloor/tests/test_stock_split.py
```diff
@@ -193,7 +193,7 @@ def _update_qty_in_location(
         # this method adds the quantity to the current quantity, so remove it
         quantity -= sum(quants.mapped("quantity"))
```
sebalix, line 194: Don't we need to also add the check `if not quantity: return` here?

shopfloor/services/cluster_picking.py
```diff
@@ -379,7 +379,7 @@ def _last_picked_line(self, picking):
         """Get the last line picked and put in a pack for this picking"""
         return fields.first(
             picking.move_line_ids.filtered(
-                lambda x: x.qty_done > 0
+                lambda x: x.qty_picked > 0
```
sebalix:
```suggestion
                lambda x: x.picked
```

shopfloor/tests/__init__.py
```diff
@@ -1,28 +1,28 @@
...
-from . import test_actions_stock
+# from . import test_menu_counters
+# from . import test_menu_contrains
+# from . import test_openapi
+# from . import test_actions_change_package_lot
+# from . import test_actions_data
+# from . import test_actions_data_detail
+# from . import test_actions_search_move_line
+# from . import test_actions_search
+# from . import test_actions_stock
```
sebalix: To restore

shopfloor/tests/test_cluster_picking_unload.py
```diff
@@ -358,8 +373,9 @@ def test_set_destination_all_error_location_move_invalid(self):
         """
         move_lines = self.move_lines
         self._set_dest_package_and_done(move_lines, self.bin1)
-        move_lines[0].move_id.location_dest_id = self.packing_a_location
-        move_lines[0].picking_id.location_dest_id = self.packing_a_location
+        # writing dest loc on all moves to avoid entering _are_all_dest_location_same
+        move_lines.move_id.location_dest_id = self.packing_a_location
```
sebalix: You probably don't need to update the destination on the move as you are updating it on the transfer below (that will update all moves)

shopfloor/tests/test_stock_split.py
mmequignon: done

shopfloor/services/cluster_picking.py
```diff
@@ -106,11 +106,11 @@ def _response_for_start_line(
             popup=popup,
         )
 
-    def _response_for_scan_destination(self, move_line, message=None, qty_done=None):
-        if qty_done is None:
+    def _response_for_scan_destination(self, move_line, message=None, qty_picked=None):
```
simahawk: why changing the signature? also, if you do it here you should do it everywhere.
On my side I'm not in favor of changing them.
``qty_done`` is a concept, not a field name, thus, for args I would keep that naming.
Plus, it will avoid headaches if we have to port things from 14 or 16.

shopfloor/tests/test_cluster_picking_scan_line.py
```diff
@@ -28,11 +28,14 @@ def _scan_line_ok(self, line, scanned, expected_qty_done=1):
         # For any barcode scanned, the quantity done is set in
         # the response data to fully done but the record is not updated.
         # We ensure the qty has not changed in the record.
-        self.assertEqual(line.qty_done, previous_qty_done)
+        self.assertEqual(line.qty_picked, previous_qty_done)
 
         self.assert_response(
             response,
             next_state="scan_destination",
+            # We kept the key as qty_done in data in order to avoid having
```
simahawk: this is not relevant IMO.
See my message above. ``qty_done`` is a concept, not a field name.

simahawk (changes requested):

sebalix (approved):

shopfloor/services/cluster_picking.py
mmequignon: Yeah, but for versions >=18, naming will be consistent.
Avoid useless headaches in the future. Drawback is headaches now.

simahawk (approved):

PR #20, opened 2025-08-05
[18.0][MIG] shopfloor: migrate `location_content_transfer` scenario

sebalix:
Depends on:
- ~https://github.com/OCA/stock-logistics-workflow/pull/2058~
- https://github.com/OCA/stock-logistics-workflow/pull/2076

Required a fix on `stock_storage_type` to make CI :green_circle: :
- https://github.com/OCA/stock-logistics-putaway/pull/12

Ref 5991

PR #35, opened 2025-08-14
[18.0][MIG] shopfloor_checkout_sync

TDu:
Depends on
- OCA/stock-logistics-workflow#1979


TDu:
Looks like the tests fail because scenario are not yet migrated in shopfloor

shopfloor/tests/test_location_content_transfer_base.py
```diff
@@ -94,7 +94,7 @@ def _simulate_pickings_selected(cls, pickings):
         """
         pickings.user_id = cls.env.uid
         for line in pickings.mapped("move_line_ids"):
-            line.qty_done = line.quantity
+            line.picked = True
```
simahawk: this should be discarded when #20 is merged

simahawk (commented):
LG
@TDu as the only scenario that is missing is content transfer, you can add #20 as a test dep

shopfloor/tests/test_location_content_transfer_base.py
TDu: done

TDu:
Removed the test dependency

simahawk (approved):

jbaudoux (approved):

PR #37, opened 2025-08-21
[18.0] shopfloor_reception: Fix tests by assigning qty_picked to reserved

grindtildeath:
Since stock.move.quantity field is not anymore computed according to the qty_picked, it highlighted an issue in the existing test where the reserved qty is not reset after being forced to 0.


simahawk:
https://github.com/OCA/stock-logistics-shopfloor/pull/39

PR #44, opened 2025-09-04
[18.0][IMP] shopfloor_reception add package's package type selection

TDu:


shopfloor_reception/actions/schema.py
```diff
@@ -0,0 +1,14 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl)
+from odoo.addons.component.core import Component
+
+
+class ShopfloorSchemaAction(Component):
+    _inherit = "shopfloor.schema.action"
+
+    def package(self, with_packaging=False):
+        res = super().package(with_packaging=with_packaging)
+        res["package_type_id:package_type"] = self._schema_dict_of(
```
simahawk: wouldn't be useful to add this in `shopfloor`?

simahawk (commented):
will you add the height editing here or you want to merge this 1st?

shopfloor_reception/actions/schema.py
TDu: It is there now, so I removed this part

shopfloor_reception/services/reception.py
```diff
@@ -1398,6 +1400,61 @@ def set_destination(
...
+            )
+            message = self._check_storage_type_valid(storage_type)
+            if not message:
+                # FIXME : this is to avoid an exception when it is required
+                selected_line.result_package_id.height = 1
+                selected_line.result_package_id.package_type_id = storage_type
+                if hasattr(selected_line, "_recompute_putaways"):
+                    # Recompute the putaway location if the module
+                    # stock_picking_putaway_recompute is installed
+                    selected_line._recompute_putaways()
```
ivantodorovich: Although I prefer glue modules, there are cases where this lazy dependency is useful. I lack context to know if this is the case.

But if you go for it, I recommend this pattern instead:

https://github.com/odoo/odoo/blob/2c02f5956f5db3137875d200133779740e19d724/addons/partner_autocomplete/models/res_partner.py#L186

```suggestion
                if self.env['ir.module.module']._get('stock_picking_putaway_recompute').state == 'installed':
                    # Recompute the putaway location if the module
                    # stock_picking_putaway_recompute is installed
                    selected_line._recompute_putaways()
```

This way, in case the `stock_picking_putaway_recompute` is refactored and somehow gets its `_recompute_putaways` method renamed or removed, you'd get a real execution error and it'd not be silently ignored.

shopfloor_reception/services/reception.py
```diff
@@ -1398,6 +1400,61 @@ def set_destination(
...
+            return self._response_for_select_dest_package(
+                picking, selected_line, message=message
+            )
+        if barcode:
+            storage_type = self.env["stock.package.type"].search(
+                [("barcode", "=", barcode)]
+            )
+            message = self._check_storage_type_valid(storage_type)
+            if not message:
+                # FIXME : this is to avoid an exception when it is required
```
ivantodorovich: For future reference, could you elaborate on this?
Which exception? When exactly? Don't hesitate to share the permalink of the related lines of code, too

That'd make it easier to re-asses if these lines are still needed in a future migration, for example.

ivantodorovich (approved):

shopfloor_reception/actions/messages.py
```diff
@@ -0,0 +1,30 @@
...
+_logger = logging.getLogger(__name__)
+
+
+class MessageAction(Component):
+    _inherit = "shopfloor.message.action"
+
+    def package_type_not_found(self):
+        return {
+            "message_type": "error",
+            "body": self.env._("The storage type could not be found"),
```
jbaudoux:
```suggestion
            "body": self.env._("The package type could not be found"),
```

shopfloor_reception/actions/messages.py
```diff
@@ -0,0 +1,30 @@
...
+        }
+
+    def package_type_changed(self):
+        return {
+            "message_type": "success",
+            "body": self.env._("The storage type was successfully changed"),
```
jbaudoux:
```suggestion
            "body": self.env._("The package type was successfully changed"),
```

shopfloor_reception/actions/messages.py
```diff
@@ -0,0 +1,30 @@
...
+        }
+
+    def storage_type_not_valid(self):
```
jbaudoux:
```suggestion
    def package_type_not_valid(self):
```

shopfloor_reception/actions/messages.py
```diff
@@ -0,0 +1,30 @@
...
+        return {
+            "message_type": "error",
+            "body": self.env._("The storage type is not valid"),
```
jbaudoux:
```suggestion
            "body": self.env._("The package type is not valid"),
```

shopfloor_reception/services/reception.py
```diff
@@ -1398,6 +1400,61 @@ def set_destination(
             return response
         return self._response_for_select_move(picking)
 
+    def set_storage_type(self, picking_id, selected_line_id, barcode=""):
+        """Set a storage type on the result package."""
```
jbaudoux:
```suggestion
    def set_package_type(self, picking_id, selected_line_id, barcode=""):
        """Set a package type on the result package."""
```

shopfloor_reception/services/reception.py
```diff
@@ -1398,6 +1400,61 @@ def set_destination(
...
+            return self._response_for_select_dest_package(
+                picking, selected_line, message=message
+            )
+        if not selected_line.exists():
+            message = self.msg_store.record_not_found()
+            return self._response_for_select_dest_package(
+                picking, selected_line, message=message
+            )
+        if barcode:
+            storage_type = self.env["stock.package.type"].search(
```
jbaudoux:
```suggestion
            package_type = self.env["stock.package.type"].search(
```

shopfloor_reception/services/reception.py
```diff
@@ -1398,6 +1400,61 @@ def set_destination(
...
+                [("barcode", "=", barcode)]
+            )
+            message = self._check_storage_type_valid(storage_type)
```
jbaudoux:
```suggestion
            message = self._check_package_type_valid(package_type)
```

shopfloor_reception/services/reception.py
```diff
@@ -1398,6 +1400,61 @@ def set_destination(
...
+            if not message:
+                # FIXME : this is to avoid an exception when it is required
+                selected_line.result_package_id.height = 1
+                selected_line.result_package_id.package_type_id = storage_type
```
jbaudoux:
```suggestion
                selected_line.result_package_id.package_type_id = package_type
```

jbaudoux (changes requested):
replace all storage type by package type
and in the filename also

jbaudoux (changes requested):
It's missing the selection when you scan a package that doesn't exist. After confirmation to create it, you should also have the package type displayed and editable

shopfloor_reception/services/reception.py
TDu: I agree, refactored. Although it was decided not to use a glue module for adding this feature.

shopfloor_reception/services/reception.py
TDu: The issue has been removed with this https://github.com/OCA/stock-logistics-putaway/pull/14

TDu:
> It's missing the selection when you scan a package that doesn't exist. After confirmation to create it, you should also have the package type displayed and editable

This is issue was caused by a Javascript error, it is fixed by https://github.com/OCA/stock-logistics-shopfloor/pull/47/commits/f4ec173846fab709090953196e635ba19652fe2b

shopfloor_reception/services/reception.py
```diff
@@ -1398,6 +1400,64 @@ def set_destination(
...
+        if message:
+            return self._response_for_select_dest_package(
+                picking, selected_line, message=message
+            )
+        if not selected_line.exists():
+            message = self.msg_store.record_not_found()
+            return self._response_for_select_dest_package(
+                picking, selected_line, message=message
+            )
+        if barcode:
```
mmequignon, line 1470: what if barcode isn't set ? we just stay on the same screen with no message?

shopfloor_reception/services/reception.py
```diff
@@ -1398,6 +1400,64 @@ def set_destination(
...
+                    selected_line._recompute_putaways()
+                message = self.msg_store.package_type_changed()
+                return self._response_for_set_destination(
+                    picking, selected_line, message=message
+                )
+        return self._response_for_set_package_type(
+            picking, selected_line, message=message
+        )
+
+    def _check_package_type_valid(self, record):
```
mmequignon, line 1495:
```suggestion
    def _check_package_type_valid(self, package_type):
```

shopfloor_reception/services/reception.py
```diff
@@ -1398,6 +1400,64 @@ def set_destination(
...
+                selected_line.result_package_id.package_type_id = package_type
+                if (
+                    self.env["ir.module.module"]
+                    ._get("stock_picking_putaway_recompute")
+                    .state
+                    == "installed"
+                ):
+                    # Recompute the putaway location if the module
+                    # stock_picking_putaway_recompute is installed
+                    selected_line._recompute_putaways()
```
mmequignon, line 1486: Should this be replaced by a hook, and be overridden in a glue module ?
I'm sorry, but I don't quite like that.
Also, this looks kinda awkward, looks like to me this should be handled with `@api,depends`
This module doesn't depend on `stock_picking_putaway_recompute` and shouldn't reference it.

shopfloor_reception/services/reception.py
```diff
@@ -1398,6 +1400,64 @@ def set_destination(
             return response
         return self._response_for_select_move(picking)
 
+    def set_package_type(self, picking_id, selected_line_id, barcode=""):
```
mmequignon:
```suggestion
    def set_package_type(self, picking_id, selected_line_id, barcode=None):
```

mmequignon (commented):
A few minor comments, and another one that is more concerning to me.

TDu:
Squashed and rebased

shopfloor_reception/services/reception.py
TDu: This endpoint handles the transition to the `set_package_type` screen, in that case the barcode would be empty.

shopfloor_reception/services/reception.py
TDu: I agree with you that this is not optimal and a glue module would be better

sebalix: 👍 it's ugly as hell :) Having a hook + glue module is the way to go.

Also, checking only `installed` state is fragile, the module could be in states `to remove` and `to upgrade`, and meanwhile the business code has to run as expected. The glue module will solve that by default.

mmequignon: I created a technical debt card for this, we can merge it.

mmequignon (approved):

shopfloor_reception/services/reception.py
mmequignon: Also, it's in prod, it works. Let's do that later (next sprint)

sebalix: Let's merge this for now, we have an ticket to cleanup this later.

PR #45, opened 2025-09-05
[18.0][FIX] shopfloor: fix migration for package type

jbaudoux:
* Delivery package (product packaging) is replaced by package type (stock.package.type)
* Having in shopfloor the packaging of the package is misleading, it serves only for setting default dimensions at reception. I removed it to prevent mistakes
* storage type renamed to package type

cc @mmequignon @TDu @simahawk

### DIFF
--- a/shopfloor/actions/data.py
+++ b/shopfloor/actions/data.py
@@ -76,15 +76,15 @@ class DataAction(Component):
         ]
 
     @ensure_model("stock.quant.package")
-    def package(self, record, picking=None, with_packaging=False, **kw):
+    def package(self, record, picking=None, with_package_type=False, **kw):
         """Return data for a stock.quant.package
 
         If a picking is given, it will include the number of lines of the package
         for the picking.
         """
         parser = self._package_parser
-        if with_packaging:
-            parser += self._package_packaging_parser
+        if with_package_type:
+            parser += self._package_package_type_parser
         data = self._jsonify(record, parser, **kw)
         qty = len(record.quant_ids)
         # handle special cases
@@ -128,7 +128,6 @@ class DataAction(Component):
             "id",
             "name",
             "shopfloor_weight:weight",
-            ("package_type_id:storage_type", ["id", "name"]),
             (
                 "quant_ids:total_quantity",
                 lambda rec, fname: sum(rec.quant_ids.mapped("quantity")),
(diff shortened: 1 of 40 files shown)


shopfloor/actions/data.py
```diff
@@ -76,15 +76,15 @@ def _picking_parser(self, **kw):
         ]
 
     @ensure_model("stock.quant.package")
-    def package(self, record, picking=None, with_packaging=False, **kw):
+    def package(self, record, picking=None, with_package_type=False, **kw):
```
TDu, line 79: This change will affect other modules like `shopfloor_reception`, no ?

shopfloor/actions/data.py
```diff
@@ -76,15 +76,15 @@ def _picking_parser(self, **kw):
...
         """Return data for a stock.quant.package
 
         If a picking is given, it will include the number of lines of the package
         for the picking.
         """
         parser = self._package_parser
-        if with_packaging:
-            parser += self._package_packaging_parser
+        if with_package_type:
+            parser += self._package_package_type_parser
```
simahawk, line 87:
```suggestion
            parser += self._package_type_parser
```

shopfloor/actions/data.py
```diff
@@ -128,17 +128,16 @@ def _package_parser(self):
...
-            ("package_type_id:storage_type", ["id", "name"]),
             (
                 "quant_ids:total_quantity",
                 lambda rec, fname: sum(rec.quant_ids.mapped("quantity")),
             ),
         ]
 
     @property
-    def _package_packaging_parser(self):
+    def _package_package_type_parser(self):
```
simahawk, line 138:
```suggestion
    def _package_type_parser(self):
```

shopfloor/actions/data.py
simahawk: definitely :)

image

TODO: check all possible overrides w/ signature change

simahawk (changes requested):
to be fixed

shopfloor/actions/data.py
jbaudoux: no, this one is right

shopfloor/actions/data.py
jbaudoux: no, this one is right

TDu:
@simahawk @jbaudoux This one impacts a lot of incoming changes...are we good to merge ?

TDu (approved):

jbaudoux:
cc @lmignon @rousseldenis for review. Then I think we need to backport this to v16

PR #46, opened 2025-09-09
[18.0][MIG] shopfloor_reception_packaging_dimension_mobile

simahawk:
TODO: Some commits must be split


TDu:
Rebased, draft status can be removed IMO, also not sure about the comment :
 
* TODO: Some commits must be split

They look good to me

shopfloor_reception/services/reception.py
```diff
@@ -181,6 +181,7 @@ def _response_for_confirm_new_package(
             next_state="confirm_new_package", data=data, message=message
         )
 
+    # TODO: is this method used somewhere?
```
TDu, line 184: It originally comes from https://github.com/OCA/wms/pull/555

But then https://github.com/OCA/wms/pull/610 removed the calls to that function.

TDu (approved):
LG, except for that function not used

mmequignon (approved):

PR #40, opened 2025-08-27
[18.0][IMP] shopfloor_reception: set package type on package assignation

TDu:
Depends on:
* https://github.com/OCA/stock-logistics-tracking/pull/49

PR #47, opened 2025-09-15
[18.0][ADD] shopfloor_reception_package_dimension

TDu:
~~Build on top of #44~~


mmequignon:
This will conflict with https://github.com/OCA/stock-logistics-shopfloor/pull/40

You add an option on the package data detail `with_package_type`.
Please, add the same on the picking data, and forward it to move lines, then on packages.
This way you can actually control that you want to display those infos or not at picking level, and this test would work:
https://github.com/OCA/stock-logistics-shopfloor/pull/40/files#diff-9767f4f3d63f298075425778b8b4bb76c320b39294131f3c103d20566503afc8R32-R41

TDu:
> This will conflict with #40
> 
> You add an option on the package data detail `with_package_type`. Please, add the same on the picking data, and forward it to move lines, then on packages. This way you can actually control that you want to display those infos or not at picking level, and this test would work: https://github.com/OCA/stock-logistics-shopfloor/pull/40/files#diff-9767f4f3d63f298075425778b8b4bb76c320b39294131f3c103d20566503afc8R32-R41

I don't think I add it, it is already there https://github.com/OCA/stock-logistics-shopfloor/blob/bac19c653652c07430b537d9baf136fca005ef5c/shopfloor/actions/schema.py#L88

jbaudoux (approved):

jbaudoux:
@Tdu Can you squash the last 3 commits regarding `shopfloor_reception_package_dimension_mobile` ?

TDu:
> @TDu Can you squash the last 3 commits regarding `shopfloor_reception_package_dimension_mobile` ?

I have squashed one, but the last one is related to another module it makes sens to me to keep it separate

TDu:
No changes, just restarting the CI

mmequignon:
bonjour

PR #49, opened 2025-09-18
 [18.0][FIX] shopfloor: format fixes for messages

rlimaeco:


shopfloor/actions/message.py
```diff
@@ -818,8 +818,8 @@ def unable_to_pick_qty(self):
     def lot_replaced_by_lot(self, old_lot, new_lot):
         return {
             "message_type": "success",
-            "body": _("Lot %(old_lot_name)s replaced by lot %(new_lot_name)s.").format(
-                old_lot.name, new_lot.name
+            "body": _("Lot {old_lot_name} replaced by lot {new_lot_name}.").format(
+                old_lot_name=old_lot.name, new_lot_name=new_lot.name
```
mmequignon:
```suggestion
            "body": _(
                "Lot %(old_lot_name)s replaced by lot %(new_lot_name)s.",
                old_lot_name=old_lot.name,
                new_lot_name=new_lot.name,
```
Please use the builtin args/kwargs.

twalter-c2c: I would also use `self.env._` instead of `_`, but that's a tiny detail, not blocking.

mmequignon: > I would also use `self.env._` instead of `_`, but that's a tiny detail, not blocking.

This is a component, there's no `self.env` here.

twalter-c2c: Right, sorry, I missed that.

twalter-c2c (approved):

mmequignon (approved):

shopfloor/actions/message.py
```diff
@@ -818,8 +818,10 @@ def unable_to_pick_qty(self):
     def lot_replaced_by_lot(self, old_lot, new_lot):
         return {
             "message_type": "success",
-            "body": _("Lot %(old_lot_name)s replaced by lot %(new_lot_name)s.").format(
-                old_lot.name, new_lot.name
+            "body": _(
```
simahawk, line 821: right below this fix there's another message that should be fixed for the same reason: `package_replaced_by_package`. Please check if there's any other message that is misusing formatting w/ `%s` w/ `.format`.

simahawk (changes requested):

shopfloor/actions/message.py
simahawk: We do have an env on components. The main reason to keep them this way is to not diverge too much from 16.0.
That's why I've not suggested to do so :wink:

simahawk (approved):

PR #51, opened 2025-09-19
[FIX] shopfloor: Avoid changing destination location after split of move line

grindtildeath:


grindtildeath:
On hold, as the issue is actually linked to `stock_dynamic_routing` module.

jbaudoux (approved):

jbaudoux:
> On hold, as the issue is actually linked to `stock_dynamic_routing` module.

Action assign recompute put away. It's a (new v18) standard behavior. That behavior has been fwd to dynamic routing but you'll get the same result without that module

PR #59, opened 2025-10-02
[18.0][FIX] shopfloor_reception: drop useless action assign

jbaudoux:
odoo is now calling the recomputation of putaway inside action assign. It cannot be called blindly anymore. I removed it from method that have no reason to make this call.

cc  @TDu

### DIFF
--- a/shopfloor/models/stock_move.py
+++ b/shopfloor/models/stock_move.py
@@ -56,15 +56,20 @@ class StockMove(models.Model):
                 == 0
             ):
                 return self.browse()
-            split_move_vals = self._split(qty_to_split)
-            split_move = self.create(split_move_vals)
-            split_move.move_line_ids = to_move
-            split_move._action_confirm(merge=False)
-            split_move._recompute_state()
-            if split_move.state != "assigned":
-                split_move._action_assign()
+            new_move_vals_list = self._split(qty_to_split)
+            confirm_dict = dict(
+                state=self.state,
+                reservation_date=self.reservation_date,
+            )
+            [d.update(confirm_dict) for d in new_move_vals_list]
+            # Only create the move, do not call _action_confirm as
+            # since Odoo 15.0, it calls _action_assign() or we only
+            # want to split
+            new_move = self.env["stock.move"].create(new_move_vals_list)
+            new_move.move_line_ids = to_move
+            new_move._recompute_state()
             self._recompute_state()
-            return split_move
+            return new_move
         return self.browse()
 
     def split_unavailable_qty(self):
--- a/shopfloor/models/stock_picking.py
+++ b/shopfloor/models/stock_picking.py
@@ -1,6 +1,6 @@
 # Copyright 2020 Camptocamp SA (http://www.camptocamp.com)
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
-from odoo import _, api, fields, models
+from odoo import api, fields, models
 
 
 class StockPicking(models.Model):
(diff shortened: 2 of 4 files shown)


TDu (approved):

shopfloor/models/stock_move.py
```diff
@@ -104,7 +104,6 @@ def _extract_in_split_order(self, default=None, backorder=False):
         self.package_level_id.picking_id = new_picking.id
         self.move_line_ids.picking_id = new_picking.id
         self.move_line_ids.package_level_id.picking_id = new_picking.id
-        self._action_assign()
```
sebalix: Why not doing:
```python
if self.state != "assigned":
    self._action_assign()
```
(I'm sure there is a reason, but it was the fix done elsewhere IIRC)

simahawk (changes requested):
needs an explanation and test update

mmequignon:
We need tests here

shopfloor/models/stock_move.py
TDu: The reason is in description of the PR

> odoo is now calling the recomputation of putaway inside action assign. It cannot be called blindly anymore. I removed it from method that have no reason to make this call.

TDu:
Rebased. The commit with the change in `shopfloor` was not needed anymore because of https://github.com/OCA/stock-logistics-shopfloor/commit/789e11ecdb9bfc4ac0d309474528045e7eb0dd92

jbaudoux:
cc @nicolas-delbovier-acsone

simahawk (approved):

simahawk:
@jbaudoux any chance to have a test that runs before and after?

jbaudoux:
@simahawk @TDu I cleanup even more, in the same way that what I did in https://github.com/OCA/stock-logistics-workflow/pull/2290/changes

TDu (approved):

shopfloor/models/stock_move.py
```diff
@@ -56,15 +56,20 @@ def split_other_move_lines(self, move_lines, intersection=False):
...
-            if split_move.state != "assigned":
-                split_move._action_assign()
+            new_move_vals_list = self._split(qty_to_split)
+            confirm_dict = dict(
+                state=self.state,
+                reservation_date=self.reservation_date,
+            )
+            [d.update(confirm_dict) for d in new_move_vals_list]
+            # Only create the move, do not call _action_confirm as
+            # since Odoo 15.0, it calls _action_assign() or we only
```
nicolas-delbovier-acsone, line 66: I guess you meant "however" (which in french is "or").

```suggestion
            # since Odoo 15.0, it calls _action_assign() however we only
```

shopfloor/models/stock_move.py
```diff
@@ -56,15 +56,20 @@ def split_other_move_lines(self, move_lines, intersection=False):
...
-            split_move._action_confirm(merge=False)
-            split_move._recompute_state()
-            if split_move.state != "assigned":
-                split_move._action_assign()
+            new_move_vals_list = self._split(qty_to_split)
+            confirm_dict = dict(
+                state=self.state,
+                reservation_date=self.reservation_date,
+            )
+            [d.update(confirm_dict) for d in new_move_vals_list]
```
lmignon: The readability is questionable IMO.... I would prefer

```suggestion
            new_move_vals_list = self._split(qty_to_split)
            for vals in new_move_vals_list:
                vals.update(
                    state=self.state,
                    reservation_date=self.reservation_date,
                )
```

PR #62, opened 2025-10-08
[FWD][IMP] shopfloor: Add a barcode parser to support multi representation

simahawk:
As a some barcodes can represent more than one kind of record,
add a helper to parse and return values.

FWD port of https://github.com/OCA/wms/pull/1001


sebalix (approved):
I guess we'll have to integrate `shopfloor_gs1` with that?

simahawk:
> I guess we'll have to integrate `shopfloor_gs1` with that?

It's already done in the related migration.

PR #64, opened 2025-10-09
[18.0][FIX] shopfloor (single_pack_transfer): check that package level is assigned

sebalix:
Reservation could not be done as expected especially since Odoo 17.0 where `product_expiry` [is filtering out the expired quants](https://github.com/odoo/odoo/blob/4d549a5712295f4164c4f725af12474ff99becbf/addons/product_expiry/models/stock_move.py#L75-L83).

This commit checks that package level is well reserved in 'Single Pack Transfer'.

It's important to get the package level assigned, so put-away are computed as expected (and also others modules like `stock_dynamic_routing` are triggered).

NOTE: with this new `product_expiry` logic, it seems impossible to move expired goods to another location (as it's impossible to reserve them). To workaround that, users could update the expiration date of a quant, or unset `use_expiration_date` on the product to move such goods.

Output when scanning an expired package:
image

### DIFF
--- a/shopfloor/actions/message.py
+++ b/shopfloor/actions/message.py
@@ -3,7 +3,7 @@
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
 import logging
 
-from odoo import _
+from odoo import _, fields
 
 from odoo.addons.component.core import Component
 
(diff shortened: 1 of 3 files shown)


jbaudoux (approved):

sebalix:
If we need to force the reservation of such expired lots, we could use this new module:
- https://github.com/OCA/stock-logistics-reservation/pull/33

PR #66, opened 2025-10-17
[18.0][IMP] shopfloor_reception: Priorize pickings the user has worked on

mmequignon:
When splitting a move line, both moves should have self.en.uid as shopfloor_user_id.
And work to be done on the picking selection screen should be sorted by user.


mmequignon:
ping @jbaudoux

sebalix (approved):

jbaudoux (approved):

PR #68, opened 2025-10-21
[18.0][IMP] shopfloor, location_content_transfer: ignore expiration date when moving goods

sebalix:
Depends on:
- https://github.com/OCA/stock-logistics-reservation/pull/33

Question: should we do the same on others scenarios? Or should we propose instead an option on all SF scenarios to bypass such reservation constraint?

### DIFF
--- a/shopfloor/__manifest__.py
+++ b/shopfloor/__manifest__.py
@@ -27,6 +27,8 @@
         #  OCA / stock-logistics-workflow
         "stock_move_line_change_lot",
         "stock_picking_progress",
+        #  OCA / stock-logistics-reservation
+        "product_expiry_assign",
         #  OCA / stock-logistics-tracking
         "stock_quant_package_dimension",
         "stock_quant_package_product_packaging",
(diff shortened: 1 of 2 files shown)


TDu:
I think an option by scenario would be good (although checkout, reception, would not make sense? ) if the configuration on the depending module stays company wide.
The question is also do we really want to add a new dependency to shofloor ? 
Anyway to refactor so it is optional ?
So many questions...

TDu:
Removed the test dependency has that pr has been merged.

simahawk:
I would add a glue module and make the flag option by menu item flag.

simahawk (changes requested):
to be fixed

mmequignon (approved):

simahawk:
> I would add a glue module and make the flag option by menu item flag.

The glue module can be one for tests only and we can leave the flag as loose dependency.

jbaudoux (changes requested):
The dependency should be in another glue module and ideally with a scenario option

PR #11, opened 2025-07-30
[18.0][MIG] shopfloor_single_product_transfer

mmequignon:
Based on:
- https://github.com/OCA/stock-logistics-workflow/pull/2031

A few FIXME here and there, some shopfloor commits were never ported to 16.0 and are missing.
A FW port PR is ready though:
- https://github.com/OCA/wms/pull/1064

However there's a few fishy things there, which I'll address later.

PR #71, opened 2025-10-22
[18.0][IMP] shopfloor checkout: capture and display validation error message

henrybackman:
When validation fails, the message was lost. Create a new message for validation errors and handle the response according to the checkout scenario.

### DIFF
--- a/shopfloor/actions/message.py
+++ b/shopfloor/actions/message.py
@@ -431,6 +431,12 @@ class MessageAction(Component):
             "body": _("Transfer {} is not available.").format(picking.name),
         }
 
+    def move_validation_failed(self, exception):
+        return {
+            "message_type": "error",
+            "body": _("Move validation failed. %s") % str(exception),
+        }
+
     def line_has_package_scan_package(self):
         return {
             "message_type": "warning",
--- a/shopfloor/services/checkout.py
+++ b/shopfloor/services/checkout.py
@@ -2,6 +2,8 @@
 # Copyright 2020 Jacques-Etienne Baudoux (BCIM) <je@bcim.be>
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
 
+import logging
+
 from werkzeug.exceptions import BadRequest
 
 from odoo import _, fields
@@ -10,6 +12,8 @@ from odoo.addons.base_rest.components.service import to_int
 from odoo.addons.component.core import Component
 
 from ..utils import to_float
+
+_logger = logging.getLogger(__name__)
 
 
 class Checkout(Component):
@@ -1558,7 +1562,21 @@ class Checkout(Component):
                 picking,
             )
         stock = self._actions_for("stock")
-        stock.validate_moves(lines_done.move_id)
+        savepoint = self._actions_for("savepoint").new()
+        try:
+            stock.validate_moves(lines_done.move_id)
+        except Exception as e:
+            savepoint.rollback()
+            _logger.error("Error while validating moves: %s", e)
+            # in case of error, go back to the location selection and pass the error
+            # to the user. This is to fix issue where user only saw a generic
+            # server error when validation failed with external API
+            return self._response_for_summary(
+                picking,
+                need_confirm=False,
+                message=self.msg_store.move_validation_failed(e),
+            )
+        savepoint.release()
         return self._response_for_select_document(
             message=self.msg_store.transfer_done_success(lines_done.picking_id)
         )
@@ -1596,7 +1614,20 @@ class Checkout(Component):
         for line in lines_done:
             line.update({"location_dest_id": scanned_location.id})
         stock = self._actions_for("stock")
-        stock.validate_moves(lines_done.move_id)
+        savepoint = self._actions_for("savepoint").new()
+        try:
+            stock.validate_moves(lines_done.move_id)
+        except Exception as e:
+            savepoint.rollback()
+            _logger.error("Error while validating moves: %s", e)
+            # in case of error, go back to the location selection and pass the error
+            # to the user. This is to fix issue where user only saw a generic
+            # server error when validation failed with external API
+            return self._response_for_select_child_location(
+                picking,
+                message=self.msg_store.move_validation_failed(e),
+            )
+        savepoint.release()
         return self._response_for_select_document(
             message=self.msg_store.transfer_done_success(lines_done.picking_id)
         )
--- a/shopfloor/tests/test_checkout_done.py
+++ b/shopfloor/tests/test_checkout_done.py
@@ -1,5 +1,13 @@
 # Copyright 2020 Camptocamp SA (http://www.camptocamp.com)
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+from unittest.mock import patch
+
+from werkzeug.exceptions import BadRequest
+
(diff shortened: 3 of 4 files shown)


shopfloor/actions/message.py
```diff
@@ -431,6 +431,12 @@ def stock_picking_not_available(self, picking):
             "body": _("Transfer {} is not available.").format(picking.name),
         }
 
+    def move_validation_failed(self, exception):
+        return {
+            "message_type": "error",
+            "body": _("Move validation failed. Message: %s") % str(exception),
+        }
```
mmequignon, line 438: Make this a little bit more useful here.
Log an error in odoo, and display the date, so when we receive a screenshot of the screen from the customer, we can easily find the traceback.

shopfloor/services/checkout.py
```diff
@@ -1596,7 +1603,13 @@ def scan_dest_location(self, picking_id, barcode):
...
             line.update({"location_dest_id": scanned_location.id})
         stock = self._actions_for("stock")
-        stock.validate_moves(lines_done.move_id)
+        try:
+            stock.validate_moves(lines_done.move_id)
+        except Exception as e:
+            return self._response_for_select_child_location(
+                picking,
+                message=self.msg_store.move_validation_failed(e),
+            )
```
mmequignon, line 1629: Please, write a meaningful comment

mmequignon (commented):
This needs a few tests

shopfloor/actions/message.py
henrybackman: Added logging

shopfloor/services/checkout.py
henrybackman: Added comment explaining the motivation

shopfloor/tests/test_checkout_done.py
```diff
@@ -25,6 +31,34 @@ def test_done_ok(self):
...
+        # line is done
+        picking.move_line_ids.write({"qty_picked": 10, "shopfloor_checkout_done": True})
+
+        # mock validation error in stock.validate_moves
+        validation_error_msg = "Validation error"
+        with patch.object(
+            StockAction, "validate_moves", side_effect=BadRequest(validation_error_msg)
+        ):
+            # mock logging to avoid error logs in test output
+            with patch("logging.Logger.error") as _:
```
mmequignon: You can use `odoo.tools.mute_logger` I believe

shopfloor/actions/message.py
```diff
@@ -431,6 +431,12 @@ def stock_picking_not_available(self, picking):
             "body": _("Transfer {} is not available.").format(picking.name),
         }
 
+    def move_validation_failed(self, exception):
+        return {
+            "message_type": "error",
+            "body": _("Move validation failed. Message: %s") % str(exception),
```
mmequignon: I'd say this isn't specific enough.
Why not `catchall_operation_failed_exception` ?
So we can use that same message everywhere we expect to get an unwanted exception.

mmequignon (commented):
Looks very good, just a few minor comments.

shopfloor/actions/message.py
henrybackman: Or do you mean it is too specific? I chose this name because going through the naming convention in the `messages.py` this name seemed to match the others in that it describes what happened. In this case we are catching issues when validating moves so this works well. 

It doesn't tell why it failed, but that is why we pass the original error message for context.

mmequignon: Yes sorry, I meant not *generic* enough.

Well that's the point, we don't know what happened. An exception was raised, that's all we know.
I'm not sure we want to implement another message next time we want to do catch exceptions when posting a picking, puting goods in a package, or whatever else.
One generic message for all this is good enough, I think.
And this might open the door for possibilities, like I don't know, a context manager catching exceptions and displaying a generic error, or a decorator on the endpoints.

shopfloor/tests/test_checkout_done.py
henrybackman: Thanks! Switched to this :+1:

shopfloor/services/checkout.py
```diff
@@ -1558,7 +1562,18 @@ def done(self, picking_id, confirmation=False):
...
         stock = self._actions_for("stock")
-        stock.validate_moves(lines_done.move_id)
+        try:
+            stock.validate_moves(lines_done.move_id)
+        except Exception as e:
+            _logger.error("Error while validating moves: %s", e)
+            # in case of error, go back to the location selection and pass the error
+            # to the user. This is to fix issue where user only saw a generic
+            # server error when validation failed with external API
+            return self._response_for_summary(
```
TDu, line 1574: In case some changes happened on validation, should we not rollback the transaction as well ?

mmequignon: I basically did this in single_product_transfer, with a decorator.
https://github.com/OCA/stock-logistics-shopfloor/pull/11/files#diff-f871450f795713d4ab1061b7117262fdee975730cb8f85f5963d06d339d7e39dR20-R45
This could go in the base module.

sebalix: For now, go for the "simplest" solution, to avoid regressions

simahawk: I agree. We can use a savepoint and roll it back in case we have the exception.

shopfloor/actions/message.py
simahawk: Let's keep it like this for now. I'd remove the "Message:" part, looks redundant.

shopfloor/tests/test_checkout_done.py
```diff
@@ -25,6 +33,31 @@ def test_done_ok(self):
...
+        picking = self._create_picking(lines=[(self.product_a, 10)])
+        self._fill_stock_for_moves(picking.move_ids, in_package=True)
+        picking.action_assign()
+        # line is done
+        picking.move_line_ids.write({"qty_picked": 10, "shopfloor_checkout_done": True})
+
+        # mock validation error in stock.validate_moves
+        validation_error_msg = "Validation error"
+        with patch.object(
+            StockAction, "validate_moves", side_effect=BadRequest(validation_error_msg)
```
simahawk, line 47: FTR this will never raise ` BadRequest`. This is raised afterwards by `base_rest` exception handler.
Here you could have a PG error or an odoo exception.
For this particular test it does not matter.

shopfloor/services/checkout.py
henrybackman: Added the savepoint

shopfloor/actions/message.py
henrybackman: Removed the "Message:" part

shopfloor/services/checkout.py
```diff
@@ -1596,7 +1614,17 @@ def scan_dest_location(self, picking_id, barcode):
         for line in lines_done:
             line.update({"location_dest_id": scanned_location.id})
         stock = self._actions_for("stock")
-        stock.validate_moves(lines_done.move_id)
+        try:
```
TDu, line 1618: We probably need the savepoint here as well. Like above

simahawk: what about this?

henrybackman: Added the savepoint to this as well

jbaudoux (approved):

PR #72, opened 2025-10-28
[FIX][18.0] shopfloor_reception: Add session recover at scan_line

mmequignon:
Add a recover feature on shopfloor_reception, because stopping the process before selecting  the destination location and after processing the full quantity would prevent the user from being able to complete the process.
If the user selected the same move again, it would create a new move line.

### DIFF
--- a/shopfloor_reception/tests/common.py
+++ b/shopfloor_reception/tests/common.py
@@ -145,6 +145,9 @@ class CommonCase(BaseCommonCase):
         for key, value in expected_message.items():
             self.assertEqual(message.get(key), value)
 
+    def assertNotMessage(self, response, expected_message):
+        self.assertMessage(response, expected_message)
+
     @classmethod
     def _get_move_ids_from_response(cls, response):
         state = response.get("next_state")
--- a/shopfloor_reception/services/reception.py
+++ b/shopfloor_reception/services/reception.py
@@ -7,7 +7,7 @@
 import pytz
 
 from odoo import fields
-from odoo.tools import float_compare
+from odoo.tools import float_compare, float_is_zero
 
 from odoo.addons.base_rest.components.service import to_int
 from odoo.addons.component.core import Component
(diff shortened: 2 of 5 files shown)


shopfloor_reception/tests/common.py
```diff
@@ -145,6 +145,9 @@ def assertMessage(self, response, expected_message):
         for key, value in expected_message.items():
             self.assertEqual(message.get(key), value)
 
+    def assertNotMessage(self, response, expected_message):
+        self.assertMessage(response, expected_message)
```
florentx, line 149: What's that?

sebalix: It's magic, like on Python 2:
```python
>>> True = False
>>> bool(True)
False
```

mmequignon: Some draft code I forgot to drop, it's unused, wanted to wrap this in an assertraises and changed my mind. It wont hurt and certainly shouldn't prevent a review ;)

simahawk (approved):
LGTM

simahawk:
@mmequignon small remark: to me it's an addition, not a fix....

mmequignon:
> @mmequignon small remark: to me it's an addition, not a fix....

https://github.com/OCA/stock-logistics-shopfloor/pull/72/files#diff-57b252eb869558e97f9dbe4859b1f4cafc7db1e67ccdaecc92d2fc38d46a71bdL272-L273
This is the fix. Reception considered a move line with a package as done, while there's another screen after set package.
I don't mind if you want this flagged this as an addition, but to me it isn't one.

TDu (approved):

TDu:
Fixed the test on `shopfloor_reception_packaging_dimension`

https://github.com/OCA/stock-logistics-shopfloor/pull/72/files#diff-8c38b290b7653815d5e79f114cfdeee3740319de2dfea88d2055056645b7fbf4

jbaudoux:
This is not working properly. I'll fix it using shopfloor_unloaded flag to know if a line was entirely processed or needs to be recovered

PR #75, opened 2025-11-13
[18.0][IMP] shopfloor_reception: imp manual_select_move

TDu:
When scanning a barcode if the resulting selected move is done, the system shows a warning and comes back to the selection.

This change implementes the same functionality but when selecting a move manually (with the checkbox)

### DIFF
--- a/shopfloor/tests/test_location_content_transfer_single.py
+++ b/shopfloor/tests/test_location_content_transfer_single.py
@@ -543,6 +543,13 @@ class LocationContentTransferSingleCase(LocationContentTransferCommonCase):
             - Only the first quantity should be transfered
             - The backorder line should be proposed after
         """
+        # Making sure that no other picking are assigned to the user
+        # Because they could be picked by the start_single at the end of the test
+        # Instead of the backorder
+        pickings = self.env["stock.picking"].search(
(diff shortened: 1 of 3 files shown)


henrybackman (approved):

shopfloor/tests/test_location_content_transfer_single.py
```diff
@@ -543,6 +543,10 @@ def test_postpone_line_ok_with_two_lines_and_view(self):
             - Only the first quantity should be transfered
             - The backorder line should be proposed after
         """
+        pickings = self.env["stock.picking"].search(
+            [("user_id", "=", self.env.uid), ("id", "!=", self.picking3.id)]
+        )
+        pickings.write({"user_id": False})
```
mmequignon, line 552: Can we have a little explanation about those changes please ?
Why are we dropping user on those pickings ?

mmequignon (approved):
LG even if I don't understand why one test was modified.

shopfloor/tests/test_location_content_transfer_single.py
TDu: I have added the explanation in the code

sebalix (approved):

PR #78, opened 2025-12-01
[18.0][FIX] shopfloor_reception: Align product_expiry dates on stock.lot

mmequignon:
As only `expiration_date` is set during reception, other dates coming from `product_expiry` should be aligned with it when set.

### DIFF
--- a/shopfloor_reception/services/reception.py
+++ b/shopfloor_reception/services/reception.py
@@ -1090,8 +1090,22 @@ class Reception(Component):
             selected_line._onchange_serial_number()
         elif expiration_date:
             selected_line.write({"expiration_date": expiration_date})
-            selected_line.lot_id.write({"expiration_date": expiration_date})
+            lot = selected_line.lot_id
+            vals = self._write_lot_expiration_values(lot, expiration_date)
+            lot.write(vals)
         return self._response_for_set_lot(picking, selected_line)
+
+    def _write_lot_expiration_values(self, lot, expiration_date):
+        # Align product_expiry dates with expiration_date
+        tmp_lot = lot.new(
+            values={"product_id": lot.product_id.id, "expiration_date": expiration_date}
+        )
+        return {
+            "expiration_date": tmp_lot.expiration_date,
+            "use_date": tmp_lot.use_date,
+            "removal_date": tmp_lot.removal_date,
+            "alert_date": tmp_lot.alert_date,
+        }
 
     def _create_lot_values(self, product, lot_name):
         return {
--- a/shopfloor_reception/tests/test_set_lot.py
+++ b/shopfloor_reception/tests/test_set_lot.py
@@ -120,6 +120,7 @@ class TestSetLot(CommonCase):
         )
 
     def test_set_expiry_date(self):
+        self.product_a.use_expiration_date = True
         # First, set the lot
         picking = self._create_picking()
         lot = self._create_lot()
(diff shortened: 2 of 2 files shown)


shopfloor_reception/services/reception.py
```diff
@@ -1090,9 +1090,20 @@ def set_lot(
...
         return self._response_for_set_lot(picking, selected_line)
 
+    def _write_lot_expiration_values(self, lot, expiration_date):
+        # Align product_expiry dates with expiration_date
+        return {
+            "expiration_date": expiration_date,
+            "use_date": expiration_date,
+            "removal_date": expiration_date,
+            "alert_date": expiration_date,
+        }
```
jbaudoux, line 1108: You need to apply the delays.
I would simulate a new lot to collect the dates from https://github.com/odoo/odoo/blob/bb82c24295ae582053a2e46a6c5a3cd9358c5e86/addons/product_expiry/models/production_lot.py#L53

jbaudoux (changes requested):

shopfloor_reception/services/reception.py
mmequignon: Good idea! I addressed your comment, could you please have a look ?
Please note that those dates are set by odoo only if product is set to use expiry date.
Do you think it's enough ?
I can think about a case where a user would set an expiry date on a lot where product isn't set to use expiry dates.
use_date, removal_date and alert_date would be False.

jbaudoux (approved):

nicolas-delbovier-acsone:
I do not get why the [compute method on lot](https://github.com/odoo/odoo/blob/bb82c24295ae582053a2e46a6c5a3cd9358c5e86/addons/product_expiry/models/production_lot.py#L42C30-L62C100) is not enough?

Is it just because you have a use case for setting an expiration date on a lot for a product with `use_expiration_date=False` ? You want to bypass the date fields being set to `False` in this case?

I am not sure setting an expiration date on products that do not require one makes sense. I am even currently working on [refactoring the "set_lot" scenario in v16](https://github.com/OCA/wms/pull/1128) and I made the date picker hidden in this case.  Could you convince me to change my mind by explaining a bit your use case?

mmequignon:
> I am not sure setting an expiration date on products that do not require one makes sense. I am even currently working on [refactoring the "set_lot" scenario in v16](https://github.com/OCA/wms/pull/1128) and I made the date picker hidden in this case. Could you convince me to change my mind by explaining a bit your use case?

Well, I don't exactly know the usecase exactly, but odoo doesn't prevent setting a expiration date on a lot linked to a product where expiration date isn't required.
That alone justifies the presence of this input field on the screen to me.

However I applied blindly what @jbaudoux but I kind of agree with you regarding setting expiration date alone, an let the compute method do its job.

nicolas-delbovier-acsone:
>  odoo doesn't prevent setting a expiration date on a lot linked to a product where expiration date isn't required.

Yes, however, Odoo's standard behavior is to clear the expiration date on lots if the product doesn't require them. I think that in order to avoid confusing bugs where data 'disappears' after being saved, we should keep the input hidden unless the product configuration actually supports it.

TDu (approved):

jbaudoux (changes requested):
Nicolas is right. It should work without your change. What is failing in your test if you revert your change?

Already managed by:
https://github.com/odoo/odoo/blob/18.0/addons/product_expiry/models/production_lot.py#L42

mmequignon:
> Nicolas is right. It should work without your change. What is failing in your test if you revert your change?
> 
> Already managed by: https://github.com/odoo/odoo/blob/18.0/addons/product_expiry/models/production_lot.py#L42

Well, apparently the issue is that the `_compute_dates` method isn't called when we write the expiration date.
I'm gonna fix that instead, but leaving this open in the meantime.

mmequignon:
@jbaudoux @nicolas-delbovier-acsone thank you for your review.
The bug was coming from odoo, trying to use `_origin` in an `@api.depends` decorated compute method, which doesn't work.
Here's the odoo/odoo pr https://github.com/odoo/odoo/pull/253318

PR #80, opened 2025-12-08
[18.0][ADD] shopfloor_reception_putaway_recompute

TDu:
Refactor the ugly hooks in shopfloor reception handling the put away, with a new glue module.

- https://github.com/OCA/stock-logistics-putaway/pull/15


shopfloor_reception/services/reception.py
```diff
@@ -1305,16 +1299,15 @@ def process_with_new_pack(self, picking_id, selected_line_id, quantity):
         self._prefill_package_type(selected_line, package)
         return self._response_for_set_destination(picking, selected_line)
 
+    def _recompute_putaway_on_line(self, line, allow_unsafe=True):
+        """Hook for the module `shopfloor_reception_putaway_recompute`."""
+        pass
```
mmequignon, line 1304: I don't understand this. This is a glue module, and it depends on stock_picking_putaway_recompute.
How is that possible that we're not able to provide any kind of behavior for this method ?
hasattr/getattr isn't an option.

TDu: That's the hook in `shopfloor_reception` module, that is why it is empty

simahawk (approved):
LGTM

jbaudoux (approved):

shopfloor_reception/services/reception.py
```diff
@@ -1305,16 +1299,15 @@ def process_with_new_pack(self, picking_id, selected_line_id, quantity):
...
         """Prefill the package type on the package before the move is done."""
         package._assign_packaging(line.product_id, line.qty_picked)
         if not package.location_id:
-            if hasattr(line, "_recompute_putaways"):
-                # Recompute the putaway location if the module
-                # stock_picking_putaway_recompute is installed
-                line.with_context(
-                    allow_unsafe_putaway_recompute=True
-                )._recompute_putaways()
+            self._recompute_putaway_on_line(line)
```
jbaudoux, line 1310: What about just doing like what odoo is doing when changing the result_package_id:
```
self.location_dest_id = default_dest_location.with_context(exclude_sml_ids=self.ids)._get_putaway_strategy(
                self.product_id, quantity=self.quantity_product_uom, package=self.result_package_id,
                packaging=self.move_id.product_packaging_id)
```

jbaudoux: My proposal is to always recompute a putaway like odoo is doing here when you set a package. As the putaway depends on the package type, it would make sense to recompute it.

See https://github.com/odoo/odoo/blob/18.0/addons/stock/models/stock_move_line.py#L250

However this is a feature missing in odoo. I think it should be outside shopfloor. And you could add it as a suggested module in shopfloor but not as a dependency.

Then you don't need here the glue module to depend on `putaway_recompute` module

WDYT?

cc @nicolas-delbovier-acsone @lmignon

lmignon: Yes it makes sense to have a module recomputing the putaway on a move line on package assign since the rule could depend on it.... It's a good idea to make it a generic addon.

PR #83, opened 2026-01-12
[18.0][FIX] shopfloor: Fix unsupported operand TypeError

grindtildeath:
As the result of stock.move.extract_and_action done is merged in a recordset of stock.picking in shopfloor.stock.action component's function validate_moves, it should not return a Boolean value in case there are no assigned move, but a False-ish empty recordset to avoid the TypeError.


TDu (commented):
👍 
Rebasing  should fix the ci

TDu (approved):

simahawk (approved):

PR #48, opened 2025-09-17
[POC][18.0] shopfloor_reception_measuring_device_mobile

mmequignon:
Includes https://github.com/OCA/stock-logistics-shopfloor/pull/46
Based on https://github.com/OCA/stock-logistics-shopfloor/pull/34

PR #86, opened 2026-01-19
[18.0][ADD] shopfloor_reception_measuring_device(_mobile)

TDu:


shopfloor_reception_measuring_device_mobile/static/src/scenario/reception_measuring_device.esm.js
```diff
@@ -0,0 +1,94 @@
...
+        <v-card-text class="details pt-0">
+            <p>The measuring device <b>{{ state.data.measuring_device.name }}</b> has been assigned for measurement of the packaging <b>{{ state.data.packaging.name }}</b>.</p>
+            <p>When done, click on OK</p>
+        </v-card-text>
+
+    </v-card>
+     <div class="button-list button-vertical-list full">
+         <v-row align="center">
+             <v-col class="text-center" cols="12">
+                 <btn-action @click="state.on_ok">OK</btn-action>
```
simahawk:
```suggestion
                 <btn-action @click="state.on_confirm_measure">OK</btn-action>
```
...or any other better naming

shopfloor_reception_measuring_device_mobile/static/src/scenario/reception_measuring_device.esm.js
```diff
@@ -0,0 +1,94 @@
...
+// Get the original template of the reception scenario
+const template = reception_scenario.component.template;
+// Add in template: the button to access the measuring device screen
+const button_placeholder = "<!-- measuring-device-placeholder -->";
+const new_template_temp = template.replace(
+    button_placeholder,
+    `
+        <v-row>
+            <v-col class="text-center" cols="12">
+                <btn-action @click="state.use_measuring_device">USE MEASURING DEVICE</btn-action>
```
simahawk: why capital case?

shopfloor_reception_measuring_device_mobile/static/src/scenario/reception_measuring_device.esm.js
```diff
@@ -0,0 +1,94 @@
...
+    new_template_temp.substring(pos_new_state);
+
+// // Extend the reception scenario with :
+// //   - the new patched template
+// //   - the js code for the new state
+const ReceptionMeasuringDevice = process_registry.extend("reception", {
+    template: new_template,
+    "methods._get_states": function () {
+        const states = _get_states.bind(this)();
+        states.set_packaging_dimension.use_measuring_device = () => {
```
simahawk: hmm this is a bit tricky to understand. 
We have a state that is called as a function of another state. Correct?
Can we find a better naming?
If this is meant to be the handler of an event I would call it `on_use_measuring_device`.
Also, note that events into states can be setup w/ just the name of the event like

```
                    events: {
                        '$event_name': '$handler',
                    },
```
then you can simply trigger an event rather then calling a specific function directly. This makes also easier to hook from the same event from outside. Please check `_state_bind_events` in the base scenario mixin.

You can also think of this as an alternative to the patch-y approach of having text hook like `<!-- measuring-device-placeholder -->`.  
You can have the base UI stuff in the base dimension module and hook through events from this module. I hope that's clear. If not, I'm available for discussion 😉

simahawk (changes requested):

simahawk:
@TDu not clear to me if we shall close https://github.com/OCA/stock-logistics-shopfloor/pull/48

shopfloor_reception_measuring_device_mobile/static/src/scenario/reception_measuring_device.esm.js
TDu: changed

shopfloor_reception_measuring_device_mobile/static/src/scenario/reception_measuring_device.esm.js
TDu: broken keyboard...fixed

shopfloor_reception_measuring_device_mobile/static/src/scenario/reception_measuring_device.esm.js
TDu: The module adds a button in one existing screen and the handler for that button.
And then a new state with its related template.
I have added some comment to make it clearer

TDu:
> @TDu not clear to me if we shall close #48

IMO it can be closed, it was a quick implementation of this but done inside the `shopfloor_reception_packaging_dimension_mobile` module.

PR #88, opened 2026-01-30
[18.0][IMP] shopfloor: cluster picking always close batch on unload

TDu:
Until now, after unloading because of a full bin (batch not fully picked), the batch is ketp open and the user is invited to keep picking the next line from the batch.
With this change the batch will be closed with the transfer done and the user is redirected to the start of the scenario.

J-E: Like in the other scenario, it is now calling validate_move instead of _action_done. Partially processed pickings are now moved to a split order. The picking type backorder strategy is now also respected.

### DIFF
--- a/shopfloor/models/stock_move.py
+++ b/shopfloor/models/stock_move.py
@@ -4,7 +4,7 @@
 from markupsafe import Markup
 
 from odoo import api, fields, models
-from odoo.tools.float_utils import float_compare
+from odoo.tools.float_utils import float_compare, float_is_zero
 
 
 class StockMove(models.Model):
@@ -18,6 +18,13 @@ class StockMove(models.Model):
             move.quantity_picked = sum(
                 ml.qty_picked for ml in move.move_line_ids if ml.picked
             )
+
+    @property
+    def has_quantity_reserved(self):
+        self.ensure_one()
+        return not float_is_zero(
+            self.quantity, precision_rounding=self.product_uom.rounding
+        )
 
     def _qty_is_satisfied(self):
         compare = float_compare(
--- a/shopfloor/models/stock_move.py
+++ b/shopfloor/models/stock_move.py
@@ -111,10 +111,9 @@ class StockMove(models.Model):
             "move_ids": [],
             "move_line_ids": [],
             "backorder_id": picking.id,
-            "batch_id": picking.batch_id.id,
         }
         data.update(dict(default or []))
-        new_picking = picking.with_context(disable_batch_sanity_check=True).copy(data)
+        new_picking = picking.copy(data)
         tmpl = (
             "<a href='#' data-oe-model='stock.picking' data-oe-id='{p_id}'>{p_name}</a>"
         )
@@ -138,6 +137,10 @@ class StockMove(models.Model):
                 if line.package_id == line.result_package_id:
                     line.result_package_id = False
 
+        # The batch cannot be set during copy as the moves have to be first
+        # extracted to the new picking in order to have a non draft state that
+        # will succeed the batch sanity check
+        new_picking.batch_id = picking.batch_id
         return new_picking
 
     def extract_and_action_done(self):
--- a/shopfloor/models/stock_move.py
+++ b/shopfloor/models/stock_move.py
@@ -111,9 +111,10 @@ class StockMove(models.Model):
             "move_ids": [],
             "move_line_ids": [],
             "backorder_id": picking.id,
+            "batch_id": picking.batch_id.id,
         }
         data.update(dict(default or []))
-        new_picking = picking.copy(data)
+        new_picking = picking.with_context(disable_batch_sanity_check=True).copy(data)
         tmpl = (
             "<a href='#' data-oe-model='stock.picking' data-oe-id='{p_id}'>{p_name}</a>"
         )
--- a/shopfloor/models/stock_picking_batch.py
+++ b/shopfloor/models/stock_picking_batch.py
@@ -39,8 +39,3 @@ class StockPickingBatch(models.Model):
 
     def _calc_weight(self, pickings):
         return sum(pickings.mapped("total_weight"))
-
-    def _sanity_check(self):
-        if self.env.context.get("disable_batch_sanity_check"):
-            return
-        return super()._sanity_check()
--- a/shopfloor/models/stock_picking_batch.py
+++ b/shopfloor/models/stock_picking_batch.py
@@ -39,3 +39,8 @@ class StockPickingBatch(models.Model):
 
     def _calc_weight(self, pickings):
         return sum(pickings.mapped("total_weight"))
+
+    def _sanity_check(self):
+        if self.env.context.get("disable_batch_sanity_check"):
+            return
+        return super()._sanity_check()
--- a/shopfloor/services/cluster_picking.py
+++ b/shopfloor/services/cluster_picking.py
@@ -77,9 +77,9 @@ class ClusterPicking(Component):
     * Declare stock out: if a good is in fact not in stock or only partially. Note
       the move lines will become unavailable or partially unavailable and will
       generate a back-order.
-    * Full bin: declaring a full bin allows to move directly to the first phase
-      (picking) to the second one (unload). The scenario will go
-      back to the first phase if some lines remain in the queue of lines to pick.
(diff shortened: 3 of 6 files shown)


shopfloor/services/cluster_picking.py
```diff
@@ -1268,29 +1269,17 @@ def _unload_end(self, batch, completion_info_popup=None):
...
-            # Unassign not validated pickings from the batch, they will be
-            # processed in another batch automatically later on
-            all_pickings.invalidate_recordset(["state"])
-            pickings_not_done = all_pickings.filtered(lambda p: p.state != "done")
-            pickings_not_done.batch_id = False
-            return self._response_for_start(
-                message=self.msg_store.batch_transfer_complete(),
-                popup=completion_info_popup,
-            )
+        all_pickings.filtered(lambda x: x.state == "assigned")._action_done()
```
jbaudoux: 1/ What happens to pickings not processed at all when you call _action_done on them?
I doubt every move is going to a backorder and the picking is marked as done. I think it better be removed from the batch and not marked as done. Can you check?

2/ What happens to pickings partially processed at all when you call _action_done on them?
What happens to the move not processed at all (picked = False)?
What happens to the move not processed partially (picked = True) but has a move line not processed (picked = False)?

shopfloor/services/cluster_picking.py
```diff
@@ -1268,29 +1269,17 @@ def _unload_end(self, batch, completion_info_popup=None):
...
+        batch.state = "done"
+        # Unassign not validated pickings from the batch, they will be
+        # processed in another batch automatically later on
+        all_pickings.invalidate_recordset(["state"])
+        pickings_not_done = all_pickings.filtered(lambda p: p.state != "done")
+        pickings_not_done.batch_id = False
```
jbaudoux: Should be managed before the call to `_action_done` as it removes their priority and set a done date.

jbaudoux (changes requested):
Can you check the different use cases are properly managed?

In case of early unload when not everything as been processed, can you log a message on the batch?
If some pickings are removed from the batch, can you post a message on the batch?

jbaudoux (approved):
I fixed the issues. Needs other reviews

rousseldenis (approved):
LGTM

nicolas-delbovier-acsone (approved):

shopfloor/models/stock_picking_batch.py
```diff
@@ -39,3 +39,8 @@ def _compute_picking_info(self):
 
     def _calc_weight(self, pickings):
         return sum(pickings.mapped("total_weight"))
+
+    def _sanity_check(self):
+        if self.env.context.get("disable_batch_sanity_check"):
```
divad1196: It can technically not work if another module raise an error before calling `super()`.
In practice, I have not seen another OCA module doing it in `_sanity_check`
-> all good, but a comment about this caveat could be good.

shopfloor/models/stock_move.py
```diff
@@ -104,9 +104,10 @@ def _extract_in_split_order(self, default=None, backorder=False):
             "move_ids": [],
             "move_line_ids": [],
             "backorder_id": picking.id,
+            "batch_id": picking.batch_id.id,
         }
         data.update(dict(default or []))
-        new_picking = picking.copy(data)
+        new_picking = picking.with_context(disable_batch_sanity_check=True).copy(data)
```
divad1196: Can we have a comment about what part of `_sanity_check` needs to be shut down?

I think it's because of
```python
if not batch.picking_ids <= batch.allowed_picking_ids:
```
in module `stock_picking_batch`.

If that's the case, then wouldn't it be better to
1. Copy without batch_id (`batch_id: None`) to keep the check
2. Make the changes so that the sanity check is respected
3. Assign the batch_id

Or keep it this way but run `_sanity_check` once everything is good.

shopfloor/services/cluster_picking.py
```diff
@@ -1244,62 +1248,85 @@ def set_destination_all(self, picking_batch_id, barcode, confirmation=None):
     def _unload_write_destination_on_lines(self, lines, location):
         stock = self._actions_for("stock")
         stock.set_destination_on_lines(lines, location)
+        if self.work.menu.unload_package_at_destination:
```
divad1196, line 1250: NOTE: while checking this field, I noticed `unload_package_at_destination_is_possible`.
This field is only used in the view, but there are no constrain between it and `unload_package_at_destination` to ensure integrity

shopfloor/services/cluster_picking.py
```diff
@@ -1244,62 +1248,85 @@ def set_destination_all(self, picking_batch_id, barcode, confirmation=None):
...
-                popup=completion_info_popup,
-            )
 
-        next_line = self._next_line_for_pick(batch)
-        if next_line:
-            return self._response_for_start_line(
-                next_line,
-                message=self.msg_store.batch_transfer_line_done(),
-                popup=completion_info_popup,
+        def has_no_quantity(picking):
```
divad1196: Since it's not a closure, it could be a method/property like the other code you did.
Any reason to make it a nested function ?

If you keep it has a property, you can probably do `.filtered("has_no_quantity")`.
As a method, you would need a `lambda p: p.has_no_quantity()`

shopfloor/services/cluster_picking.py
```diff
@@ -1244,62 +1248,85 @@ def set_destination_all(self, picking_batch_id, barcode, confirmation=None):
...
-            # Unassign not validated pickings from the batch, they will be
-            # processed in another batch automatically later on
-            all_pickings.invalidate_recordset(["state"])
-            pickings_not_done = all_pickings.filtered(lambda p: p.state != "done")
-            pickings_not_done.batch_id = False
-            return self._response_for_start(
-                message=self.msg_store.batch_transfer_complete(),
-                popup=completion_info_popup,
+
+        empty_pickings = batch.picking_ids.filtered(has_no_quantity)
```
divad1196: We never check that batch is defined

shopfloor/services/cluster_picking.py
```diff
@@ -1244,62 +1248,85 @@ def set_destination_all(self, picking_batch_id, barcode, confirmation=None):
...
+                        ml.display_name,
+                    )
+                )
+            if not ml.shopfloor_unloaded:
+                # A move line is not unloaded, exit
+                return
+        stock = self._actions_for("stock")
+        for move in picking.move_ids:
+            move.split_other_move_lines(move.move_line_ids.filtered("picked"))
+        # remove assigned non picked moves
```
divad1196, line 1279: Filter technically work
- `not m.state == "assigned"` -> either "not assigned`
- `or m.picked` -> if assigned, then must be picked

But hard to read as it is.

I would recommend 2 steps
```python
moves_to_ignore = picking.move_ids.filtered(
            lambda m: m.state == "assigned" and not m.picked
)
moves_to_validate = picking.move_ids - moves_to_ignore
```

Or using negation (less clean IMO)
```python
        moves_to_validate = picking.move_ids.filtered(
            lambda m: not (m.state == "assigned" and not m.picked)
        )
```

divad1196 (commented):
(sorry I forgot to submit my review sooner)

shopfloor/models/stock_move.py
jbaudoux: Thanks for the advice. Indeed, if I set the batch on the new picking after having inserted the moves, it will be in the right state for the sanity check. A lot better :)

shopfloor/services/cluster_picking.py
jbaudoux: it's tested by the caller

shopfloor/services/cluster_picking.py
jbaudoux: coming from a copy/paste. I rewrite it

shopfloor/services/cluster_picking.py
jbaudoux: I thought comment was enough but let's make it more readable using negation

shopfloor/services/cluster_picking.py
jbaudoux: Indeed but I keep this outside the scope of this PR

divad1196 (approved):

PR #91, opened 2026-02-27
[18.0][FIX] FakeModel

jbaudoux:
cc @TDu @grindtildeath @rousseldenis

### DIFF
--- a/shopfloor/tests/test_checkout_list_package_type.py
+++ b/shopfloor/tests/test_checkout_list_package_type.py
@@ -8,53 +8,33 @@ from .test_checkout_select_package_base import CheckoutSelectPackageMixin
 
 # pylint: disable=missing-return
 class CheckoutListDeliveryPackagingCase(CheckoutCommonCase, CheckoutSelectPackageMixin):
-    @classmethod
-    def setUpClass(cls):
-        try:
-            super().setUpClass()
-        except BaseException:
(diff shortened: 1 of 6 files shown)


rousseldenis (approved):

TDu (approved):

shopfloor/tests/test_checkout_list_package_type.py
```diff
@@ -8,53 +8,33 @@
...
-        super().tearDownClass()
-
-    @classmethod
-    def setUpClassBaseData(cls, *args, **kwargs):
-        super().setUpClassBaseData(*args, **kwargs)
-        cls._load_test_models()
-        cls.carrier = cls.env["delivery.carrier"].search([], limit=1)
-        cls.carrier.sudo().delivery_type = "test"
-        cls.picking = cls._create_picking(
+        self.carrier = self.env["delivery.carrier"].search([], limit=1)
```
simahawk, line 20: I know this has been done in a rush, it's fine. However, converting _everything_ to run at `setUp` blindly is not a good idea because every record will be re-created for each an every test.
I don't know if it is required in this case, I just want to mention to be careful next time, if we have time 😉 

Late on the party... but yeah 😅

jbaudoux: yeah, create a PR to improve it :)

PR #93, opened 2026-03-06
[18.0][IMP] shopfloor: add check for package carrier

henrybackman:
- Added check in shopfloor checkout for the `package_carrier_id` which allows to restrict packages to dedicated carriers, instead of relying only on the carrier type (`package_carrier_type`)

TODO:
Where to introduce the package_carrier_id field is not yet clear and need to be decided. For now temporarily in shopfloor for testing and demonstration

### DIFF
--- a/shopfloor/services/checkout.py
+++ b/shopfloor/services/checkout.py
@@ -1175,18 +1175,20 @@ class Checkout(Component):
         self, picking, selected_lines, package_type, **kw
     ):
         carrier = self._get_carrier(picking)
+        message = None
+        is_valid = True
         if carrier:
-            # Validate against carrier
-            is_valid = self._package_type_good_for_carrier(package_type, carrier)
-        else:
-            is_valid = True
-        if carrier and not is_valid:
+            if not self._package_type_good_for_carrier(package_type, carrier):
+                is_valid = False
+                message = self.msg_store.package_type_invalid_for_carrier(package_type, carrier)
+            elif not self._package_carrier_matches_picking_carrier(package_type, carrier):
+                is_valid = False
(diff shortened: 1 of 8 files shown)


jbaudoux (commented):
Implement it in the backend first so that when you launch the 'put in pack' action, you can only select a valid package type. For this purpose, make a module in the delivery-carrier repo. That module could rely on `https://github.com/OCA/delivery-carrier/tree/18.0/stock_picking_delivery_package_type_domain` to extend the computed domain.

Shopfloor should display the same list of package types that the backend shows.

shopfloor/services/checkout.py
```diff
@@ -1175,18 +1175,20 @@ def _scan_package_action_from_package_type(
...
+                is_valid = False
+                message = self.msg_store.package_type_carrier_mismatch(package_type, carrier)
+        if not is_valid:
             return self._response_for_select_package(
                 picking,
                 selected_lines,
-                message=self.msg_store.package_type_invalid_for_carrier(
-                    package_type, carrier
-                ),
+                message=message,
```
jbaudoux, line 1191: No need to change this.
The check is performed by `actions.package_type_valid_for_carrier(package_type, carrier)` which should know what is the list of valid package types. That action should be refactored to rely on the backend to provide the right list. I think the list of valid package types is displayed on the screen, so there should be already something. Otherwise, look at the module `shopfloor_cluster_picking_repack` how to leverage `stock_picking_delivery_package_type_domain`

henrybackman:
Replaced by:
- https://github.com/OCA/stock-logistics-shopfloor/pull/102

PR #94, opened 2026-03-09
[18.0][FIX] shopfloor: Propagate lot for inventory action

grindtildeath:
As the lot was not propagated from the move line, no quant without lot could be found, and we ended up creating a quant without lot for a product tracked by lot.

### DIFF
--- a/shopfloor/actions/inventory.py
+++ b/shopfloor/actions/inventory.py
@@ -22,9 +22,9 @@ class InventoryAction(Component):
         # see comment in models/stock_inventory.py
         return self.env["stock.quant"].with_context(_sf_inventory=True)
 
-    def create_draft_check_empty(self, location, product, ref=None):
+    def create_draft_check_empty(self, location, product, ref=None, lot=None):
         """Create a draft inventory for a product with a zero quantity"""
-        return self._create_draft_inventory(location, product)
+        return self._create_draft_inventory(location, product, lot=lot)
 
     def _inventory_exists(self, location, product, package=None, lot=None):
         """Return if an inventory for location and product exist"""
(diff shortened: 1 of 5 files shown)


henrybackman (approved):

jbaudoux (approved):

shopfloor/actions/inventory.py
```diff
@@ -64,17 +64,23 @@ def _create_draft_inventory(self, location, product, package=None, lot=None):
                         # Set a user to prevent the zero quant cleanup
                         "user_id": self.env.user.id,
                         "inventory_quantity": 0,
+                        "inventory_quantity_set": True,
                         "inventory_date": fields.Date.today(),
                     }
                 )
             return quants
         else:
+            # FIXME: we should probably not get there as every move line
```
TDu, line 73: Maybe we get here if, for the system a location is empty (no existing quant) but for the user in the warehouse there is some goods in the location. In some scenario the user is asked to confirm a location is empty !

shopfloor/actions/inventory.py
```diff
@@ -64,17 +64,23 @@ def _create_draft_inventory(self, location, product, package=None, lot=None):
...
+            #  in stock, we must have a negative reserved quantity
             return self.inventory_model.sudo().create(
                 {
+                    # Set a user to prevent the zero quant cleanup
+                    "user_id": self.env.user.id,
                     "location_id": location.id,
                     "product_id": product.id,
                     "lot_id": lot.id if lot else False,
-                    "inventory_quantity": 1,
+                    "inventory_quantity": 0,
```
TDu, line 83: From my comment above. The quantity was set to 1 because the user did not confirm the location is empty !
I don't think this should be set to 0.

shopfloor/actions/inventory.py
grindtildeath: I left this comment, because AFAICS, from shopfloor to request an inventory, you need a move line, and if there is a move line, there must be a quant in the source location..

shopfloor/actions/inventory.py
grindtildeath: We checked this with @jbaudoux and the idea is to reflect what is done above, and IIRC our assumption was that it was set to 1 to avoid 0 quant cleanup, but by setting a user on it we achieve the same.

Anyway, as I wrote above we should never get to this create, as it happened only because the lot was not propagated, and we should get to the write instead.

TDu (approved):

grindtildeath:
@jbaudoux merge? :pray:

PR #95, opened 2026-03-11
[18.0][FIX] shopfloor: filter out internal packs in checkout scenario

grindtildeath:


jbaudoux (commented):
Looks good to me. Just remove the comment :)
That concept of package_carrier_type is already used in this checkout service at other places:
* https://github.com/OCA/stock-logistics-shopfloor/blob/18.0/shopfloor/services/checkout.py#L1212
* https://github.com/OCA/stock-logistics-shopfloor/blob/18.0/shopfloor/services/checkout.py#L1216
* https://github.com/OCA/stock-logistics-shopfloor/blob/18.0/shopfloor/services/checkout.py#L1427

So let's keep it simple

shopfloor/services/checkout.py
```diff
@@ -129,8 +129,12 @@ def _data_for_packing_info(self, picking):
         return ""
 
     def _response_for_select_dest_package(self, picking, move_lines, message=None):
+        # TODO: Extract filter function to a customizable parameter since it is almost
+        #  the same logic than what is defined in stock.picking._get_packages_to_pack
+        #  but here we want only delivery package and not internal ones.
         packages = picking.mapped("move_line_ids.result_package_id").filtered(
-            "package_type_id"
+            lambda pack: pack.package_type_id
+            and not pack.package_type_id.package_carrier_type
```
jbaudoux: You forgot to test if there is a carrier. If no carrier, then internal package, otherwise no internal package
```suggestion
            lambda pack, picking=picking:
            pack.package_type_id.package_carrier_type
            if picking.carrier_id
            else (
               not pack.package_type_id
               or not pack.package_type_id.package_carrier_type
            )
```

shopfloor/services/checkout.py
```diff
@@ -130,7 +130,13 @@ def _data_for_packing_info(self, picking):
...
     def _response_for_select_dest_package(self, picking, move_lines, message=None):
         packages = picking.mapped("move_line_ids.result_package_id").filtered(
-            "package_type_id"
+            lambda pack, picking=picking: pack.package_type_id.package_carrier_type
+            and pack.package_type_id.package_carrier_type != "none"
+            if picking.carrier_id
+            else (
+                pack.package_type_id
+                and pack.package_type_id.package_carrier_type == "none"
+            )
```
jbaudoux: When there is a carrier, there must be a package_carrier_type ("none" or another value).
When there is no carrier, we don't know if it is an internal package or a customer package as the carrier is not mandatory but we are sure we don't want a package type bound to a carrier.

Using https://github.com/OCA/stock-logistics-tracking/blob/18.0/internal_stock_quant_package/models/stock_quant_package.py#L10

```suggestion
            lambda pack, picking=picking:
            not pack.is_internal
            if picking.carrier_id
            else (
                not pack.package_type_id
                or pack.is_internal
            )
```

where

```
record.is_internal = (
                record.package_type_id
                and not record.package_type_id.package_carrier_type
            )
```

or without the dependency but aligned with the definition:
```suggestion
            lambda pack, picking=picking: 
            (not pack.package_type_id or pack.package_type_id.package_carrier_type)
            if picking.carrier_id
            else (
                not pack.package_type_id
                or not pack.package_type_id.package_carrier_type
            )
```

jbaudoux (changes requested):

shopfloor/services/checkout.py
grindtildeath: In your suggestion above (that doesn't use the dependency) the same condition `not pack.package_type_id` appears both in `if picking.carrier_id`  and `else` :confused:

jbaudoux: Indeed you can factorize

shopfloor/models/stock_quant_package.py
```diff
@@ -99,3 +99,11 @@ def _move_package_quant_move_values(self, quant, location, quantity):
             "package_id": quant.package_id.id,
             "owner_id": quant.owner_id.id,
         }
+
+    def _filter_for_picking_carrier(self, picking=None):
```
simahawk: would be nice to have a docstring :)

simahawk: discussed over daily: we keep it as it is

shopfloor/models/stock_quant_package.py
```diff
@@ -99,3 +99,19 @@ def _move_package_quant_move_values(self, quant, location, quantity):
             "package_id": quant.package_id.id,
             "owner_id": quant.owner_id.id,
         }
+
+    def _filter_for_picking_carrier_checkout(self, picking=None):
+        """Filter packs to be displayed in shopfloor checkout scenario
```
jbaudoux: Can you clarify if it is for displaying the packs to take from or the packs that are already done

jbaudoux (commented):
Can you squash the fixups

jbaudoux (approved):

henrybackman (approved):

shopfloor/tests/test_checkout_list_package.py
```diff
@@ -112,7 +111,8 @@ def setUpClassBaseData(cls):
                 (cls.product_b, 10),
                 (cls.product_c, 10),
                 (cls.product_d, 10),
-            ]
+            ],
+            # **{"carrier_id": cls.carrier}
```
TDu: to remove

TDu (approved):

shopfloor/tests/test_checkout_list_package.py
```diff
@@ -52,15 +52,14 @@ def test_list_dest_package_ok(self):
                 (self.product_b, 10),
                 (self.product_c, 10),
                 (self.product_d, 10),
-            ]
+            ],
+            **{"carrier_id": self.carrier},
```
mmequignon, line 56:
```suggestion
            carrier_id=self.carrier ,
```

shopfloor/models/stock_quant_package.py
```diff
@@ -99,3 +99,19 @@ def _move_package_quant_move_values(self, quant, location, quantity):
...
+          has its carrier type defined
+        - Carrier is not defined on the picking: Keep packs without a package type
+          or with a package type that does not have a carrier type defined
+        """
+        self.ensure_one()
+        return (
+            not self.package_type_id or not self.package_type_id.package_carrier_type
+            if not picking.carrier_id
+            else (self.package_type_id and self.package_type_id.package_carrier_type)
+        )
```
mmequignon, line 117: great one-liner :D

mmequignon (approved):
LG overall

simahawk:
rebased

PR #101, opened 2026-04-14
shopfloor: Fix non deterministic unittests

mmequignon:
using a location used by potentially any other module creates varying data for that location.
ping @jbaudoux

### DIFF
--- a/shopfloor/tests/test_actions_data.py
+++ b/shopfloor/tests/test_actions_data.py
@@ -1,6 +1,7 @@
 # Copyright 2020 Camptocamp SA (http://www.camptocamp.com)
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
 # pylint: disable=missing-return
+
 from markupsafe import Markup
 
 from .common import PickingBatchMixin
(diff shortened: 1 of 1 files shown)


shopfloor/tests/test_actions_data.py
```diff
@@ -41,21 +42,54 @@ def test_data_location_no_barcode(self):
...
         }
         self.assertDictEqual(data, expected)
+        # Move is confirmed, but not assigned: no move line to count qty from
+        move._action_confirm()
+        data = self.data.location(location, with_operation_progress=True)
+        self.assertDictEqual(data, expected)
+        # Move is assigned, but no qty in stock: no qty to process
+        move._action_assign()
+        data = self.data.location(location, with_operation_progress=True)
+        self.assertDictEqual(data, expected)
```
jbaudoux: action_confirm calls action_assign, so you can reduce that to one test

jbaudoux (approved):

shopfloor/tests/test_actions_data.py
mmequignon: Done

PR #102, opened 2026-04-16
[18.0][IMP] shopfloor: filter packages based on dedicated carriers

henrybackman:
Rely on backend choose.delivery.package wizard to collect the list of valid package types.

Depends on:
* https://github.com/OCA/delivery-carrier/pull/1214

that allows to limit the list of package types based on the carrier (and not only the carrier type).

The new config flag `filter_package_type_on_put_in_pack` in `stock_picking_delivery_package_type_domain` needs to be set to keep filtering package types by carrier.

### DIFF
--- a/shopfloor/__manifest__.py
+++ b/shopfloor/__manifest__.py
@@ -44,6 +44,8 @@
         "product_packaging_level",
         #  OCA / delivery
         "stock_picking_delivery_link",
+        #  OCA / delivery-carrier
+        "stock_picking_delivery_package_type_domain",
         # TODO v18: new dependency due to ``available_carriers`` that needs a SO.
         # see picking_form.py
         "sale_stock",
--- a/shopfloor/actions/packing.py
+++ b/shopfloor/actions/packing.py
@@ -11,15 +11,42 @@ class PackingAction(Component):
     _inherit = "shopfloor.process.action"
     _usage = "packing"
 
-    def package_type_valid_for_carrier(self, package_type, carrier):
-        return package_type.package_carrier_type in (
-            "none",
-            carrier.delivery_type,
-        )
+    def _delivery_type_for_carrier(self, carrier):
+        # "fixed" and "base_on_rule" are pricing rules, not real carrier integrations,
+        # so they have no package type restrictions, treat them as no carrier.
+        # See: odoo/src/addons/stock_delivery/models/stock_picking.py line 146-147
+        if not carrier:
+            return False
+        elif carrier.delivery_type in ("fixed", "base_on_rule", "pricelist"):
+            return "none"
+        return carrier.delivery_type
+
+    def _get_carrier(self, picking):
+        return picking.ship_carrier_id or picking.carrier_id
+
+    def _get_package_type_domain(self, picking):
+        carrier = self._get_carrier(picking)
+        wizard_obj = self.env["choose.delivery.package"]
+        delivery_type = self._delivery_type_for_carrier(carrier)
+        wizard = wizard_obj.with_context(
+            current_package_carrier_type=delivery_type
+        ).new({"picking_id": picking.id})
+        return wizard.package_type_domain
+
+    def package_type_valid_for_carrier(self, package_type, picking):
+        carrier = self._get_carrier(picking)
+        if not carrier:
+            return True
+        domain = self._get_package_type_domain(picking)
+        return bool(package_type.filtered_domain(domain))
 
     def create_delivery_package(self, carrier):
         default_package_type = self._get_default_package_type(carrier)
         return self.create_package_from_package_type(default_package_type)
+
+    def available_package_types_for_picking(self, picking, order="name"):
+        domain = self._get_package_type_domain(picking)
+        return self.env["stock.package.type"].search(domain, order=order)
 
     def _get_default_package_type(self, carrier):
         # TODO: refactor `delivery_[carrier_name]` modules
--- a/shopfloor/services/checkout.py
+++ b/shopfloor/services/checkout.py
@@ -725,17 +725,11 @@ class Checkout(Component):
             4. assign new package and skip `select_package` state
 
         """
-        carrier = self._get_carrier(picking)
-        if carrier:
-            # Validate against carrier
-            is_valid = self._package_type_good_for_carrier(package_type, carrier)
-        else:
-            is_valid = True
-        if carrier and not is_valid:
+        if not self._package_type_valid_for_carrier(picking, package_type):
             return self._response_for_select_line(
                 picking,
                 message=self.msg_store.package_type_invalid_for_carrier(
-                    package_type, carrier
+                    package_type, self._get_carrier(picking)
                 ),
             )
         message = None
@@ -1178,18 +1172,12 @@ class Checkout(Component):
     def _scan_package_action_from_package_type(
         self, picking, selected_lines, package_type, **kw
     ):
-        carrier = self._get_carrier(picking)
-        if carrier:
-            # Validate against carrier
-            is_valid = self._package_type_good_for_carrier(package_type, carrier)
-        else:
-            is_valid = True
-        if carrier and not is_valid:
+        if not self._package_type_valid_for_carrier(picking, package_type):
             return self._response_for_select_package(
                 picking,
                 selected_lines,
                 message=self.msg_store.package_type_invalid_for_carrier(
-                    package_type, carrier
+                    package_type, self._get_carrier(picking)
                 ),
             )
         return self._create_and_assign_new_package_type(
@@ -1204,22 +1192,13 @@ class Checkout(Component):
     def _get_carrier(self, picking):
         return picking.ship_carrier_id or picking.carrier_id
 
-    def _package_type_good_for_carrier(self, package_type, carrier):
+    def _package_type_valid_for_carrier(self, picking, package_type):
         actions = self._actions_for("packing")
-        return actions.package_type_valid_for_carrier(package_type, carrier)
+        return actions.package_type_valid_for_carrier(package_type, picking)
 
     def _get_available_package_type(self, picking):
-        model = self.env["stock.package.type"]
-        carrier = picking.ship_carrier_id or picking.carrier_id
-        if not carrier:
-            return model.search(
-                [("package_carrier_type", "=", False)],
-                order="name",
-            )
-        return model.search(
-            [("package_carrier_type", "=", carrier.delivery_type or "none")],
-            order="name",
-        )
+        actions = self._actions_for("packing")
+        return actions.available_package_types_for_picking(picking)
 
     def list_package_type(self, picking_id, selected_line_ids):
         """List available package type for given picking.
--- a/shopfloor/services/zone_picking.py
+++ b/shopfloor/services/zone_picking.py
@@ -1304,7 +1304,7 @@ class ZonePicking(Component):
         carrier = picking.ship_carrier_id or picking.carrier_id
         if carrier:
             actions = self._actions_for("packing")
-            if actions.package_type_valid_for_carrier(package.package_type_id, carrier):
+            if actions.package_type_valid_for_carrier(package.package_type_id, picking):
                 good_for_packing = True
             else:
                 message = self.msg_store.package_type_invalid_for_carrier(
--- a/shopfloor/tests/test_checkout_scan_package_action.py
+++ b/shopfloor/tests/test_checkout_scan_package_action.py
@@ -3,11 +3,29 @@
 from itertools import product
 from unittest import mock
 
+from odoo_test_helper import FakeModelLoader
+
 from .test_checkout_base import CheckoutCommonCase
 from .test_checkout_select_package_base import CheckoutSelectPackageMixin
 
 
 class CheckoutScanPackageActionCase(CheckoutCommonCase, CheckoutSelectPackageMixin):
+    def setUp(self):
+        super().setUp()
+        self.loader = FakeModelLoader(self.env, self.__module__)
+        self.loader.backup_registry()
+
+        from .models import DeliveryCarrierTest, StockPackageType
+
+        self.loader.update_registry((DeliveryCarrierTest, StockPackageType))
+
+        self.carrier = self.env["delivery.carrier"].search([], limit=1)
+        self.carrier.sudo().delivery_type = "test"
+
+    def tearDown(self):
+        self.loader.restore_registry()
+        super().tearDown()
+
     def _test_select_product(
         self, barcode_func, origin_qty_func, expected_qty_func, in_lot=False
     ):
@@ -381,23 +399,18 @@ class CheckoutScanPackageActionCase(CheckoutCommonCase, CheckoutSelectPackageMix
                 }
             )
         )
-        # Delivery type and package_carrier_type values
-        # depend on specific implementations that we don't have as dependency.
-        # What is important here is to simulate their value when mismatching.
-        mock1 = mock.patch.object(
-            type(packaging),
-            "package_carrier_type",
-            new_callable=mock.PropertyMock,
-        )
-        mock2 = mock.patch.object(
-            type(picking.carrier_id),
-            "delivery_type",
-            new_callable=mock.PropertyMock,
-        )
-        with mock1 as mocked_package_carrier_type, mock2 as mocked_delivery_type:
-            # Not matching at all -> bad
-            mocked_package_carrier_type.return_value = "DHL"
-            mocked_delivery_type.return_value = "UPS"
+        packing_action = type(self.service._actions_for("packing"))
+        with mock.patch.object(
+            packing_action,
+            "_get_package_type_domain",
+            side_effect=[
+                # first call: domain excludes packaging -> bad
+                [("id", "!=", packaging.id)],
+                # second call: domain includes packaging -> good
+                [("id", "=", packaging.id)],
+            ],
+        ):
+            # not in wizard domain -> bad
             response = self.service.dispatch(
                 "scan_package_action",
                 params={
@@ -414,8 +427,7 @@ class CheckoutScanPackageActionCase(CheckoutCommonCase, CheckoutSelectPackageMix
                     packaging, picking.carrier_id
                 ),
             )
-            # No carrier type set on the packaging -> good
-            mocked_package_carrier_type.return_value = "none"
+            # in wizard domain -> good
             response = self.service.dispatch(
                 "scan_package_action",
                 params={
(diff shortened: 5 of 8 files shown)


shopfloor/services/checkout.py
```diff
@@ -1206,16 +1206,24 @@ def _package_type_good_for_carrier(self, package_type, carrier):
...
         model = self.env["stock.package.type"]
-        carrier = picking.ship_carrier_id or picking.carrier_id
+        carrier = self._get_carrier(picking)
+        domain = []
         if not carrier:
-            return model.search(
-                [("package_carrier_type", "=", False)],
-                order="name",
+            domain = [("package_carrier_type", "=", False)]
+            if "package_carrier_id" in model._fields:
```
TDu: Not needed if the module `stock_picking_delivery_package_type_carrier` is added in the dependencies

shopfloor/services/checkout.py
```diff
@@ -1206,16 +1206,24 @@ def _package_type_good_for_carrier(self, package_type, carrier):
...
+                domain.append(("package_carrier_id", "=", False))
+            return model.search(domain, order="name")
+
+        domain = [("package_carrier_type", "=", carrier.delivery_type or "none")]
+        if "package_carrier_id" in model._fields:
```
TDu:
```suggestion

```

TDu (commented):
Please add the depending module as a temporary ref cf. https://github.com/OCA/maintainer-tools/wiki/Use-temporary-reference%28s%29-to-another-pull-request%28s%29

shopfloor/services/checkout.py
```diff
@@ -1206,16 +1206,22 @@ def _package_type_good_for_carrier(self, package_type, carrier):
...
-            order="name",
-        )
+            domain = [
+                ("package_carrier_type", "=", False),
+                ("package_carrier_id", "=", False),
+            ]
+            return model.search(domain, order="name")
+
+        domain = [
+            ("package_carrier_type", "=", carrier.delivery_type or "none"),
```
mmequignon:
```suggestion
            ("package_carrier_type", "in", (carrier.delivery_type, "none")),
```

mmequignon (changes requested):

shopfloor/tests/test_checkout_change_package_type.py
```diff
@@ -82,6 +82,29 @@ def test_list_package_type_ok(self):
...
+        self.assert_response(
+            response,
+            next_state="change_package_type",
+            data={
+                "picking": self._picking_summary_data(self.picking),
+                "package": self._package_data(self.package, self.picking),
+                "package_type": [
+                    self._package_type_data(package_type)
+                    for package_type in self.package_type_inner_box
+                    + self.package_type_pallet
```
mmequignon:
```suggestion
                    self._package_type_data(self.package_type_inner_box),
                    self._package_type_data(self.package_type_pallet),
```

mmequignon: Why are you expecting package_type_pallet ? Is it configured with that carrier already ?
If so, care to write a comment explaining why you expect that value ?

shopfloor/actions/packing.py
```diff
@@ -12,10 +12,16 @@ class PackingAction(Component):
...
-        return package_type.package_carrier_type in (
-            "none",
-            carrier.delivery_type,
-        )
+        # validate carrier type
+        if package_type.package_carrier_type not in ("none", carrier.delivery_type):
+            return False
+        # validate specific carrier if set
+        if (
+            package_type.package_carrier_id
```
jbaudoux: Use the put in pack wizard to know the list of allowed package types

shopfloor/__manifest__.py
```diff
@@ -48,6 +48,8 @@
         "product_packaging_level",
         #  OCA / delivery
         "stock_picking_delivery_link",
+        #  OCA / delivery-carrier
+        "stock_picking_delivery_package_type_carrier",
```
jbaudoux: Do not add such dependency that complexity the module

jbaudoux (changes requested):
You need to remove all that complexity from shopfloor and rely on standard that can provide you the list of valid package types.
The new module already extend the standard to limit the list. No need to reproduce again the complexity here.

shopfloor/actions/packing.py
```diff
@@ -11,16 +11,34 @@ class PackingAction(Component):
...
+        delivery_type = self._delivery_type_for_carrier(
+            carrier or picking.ship_carrier_id or picking.carrier_id
         )
+        wizard = wizard_obj.with_context(
+            current_package_carrier_type=delivery_type
+        ).new({"picking_id": picking.id})
+        return wizard.package_type_domain
+
+    def package_type_valid_for_carrier(self, package_type, carrier, picking):
+        domain = [("id", "=", package_type.id)]
```
jbaudoux: You can simply use filtered_domain

shopfloor/services/checkout.py
```diff
@@ -724,7 +724,9 @@ def _select_lines_from_package_type(
         carrier = self._get_carrier(picking)
         if carrier:
             # Validate against carrier
-            is_valid = self._package_type_good_for_carrier(package_type, carrier)
+            is_valid = self._package_type_good_for_carrier(
```
jbaudoux: Either you rely on get_carrier outside the method or you remove carrier argument and call it inside, but don't repeat the content of get_carrier inside the called method

shopfloor/services/checkout.py
```diff
@@ -1177,7 +1179,9 @@ def _scan_package_action_from_package_type(
...
```
jbaudoux: Same here. I would move get_carrier inside the method as it's repeated and needed for the test

shopfloor/actions/packing.py
```diff
@@ -11,16 +11,34 @@ class PackingAction(Component):
     _inherit = "shopfloor.process.action"
     _usage = "packing"
 
-    def package_type_valid_for_carrier(self, package_type, carrier):
-        return package_type.package_carrier_type in (
-            "none",
-            carrier.delivery_type,
+    def _delivery_type_for_carrier(self, carrier):
+        if not carrier or carrier.delivery_type in ("fixed", False):
```
jbaudoux: I haven't dug into it but can you clarify why fixed is excluded?

shopfloor/README.rst
```diff
@@ -37,12 +37,12 @@ operations.
 
 The application supports scenarios, to relate to Operation Types:
 
-- Cluster Picking
-- Zone Picking
-- Checkout/Packing
-- Delivery
-- Location Content Transfer
-- Single Pack Transfer
+-  Cluster Picking
```
jbaudoux: All those added spaces looks weird

shopfloor/__manifest__.py
```diff
@@ -48,6 +48,8 @@
         "product_packaging_level",
         #  OCA / delivery
         "stock_picking_delivery_link",
+        #  OCA / delivery-carrier
+        "stock_picking_delivery_package_type_domain",
```
jbaudoux, line 48: Still needed?

shopfloor/static/description/index.html
```diff
@@ -372,9 +372,9 @@ <h1>Shopfloor</h1>
...
...p><a class="reference external image-reference" href="https://odoo-community.org/page/development-status"><img alt="Beta" src="https://img.shields.io/badge/maturity-Beta-yellow.png" /></a> <a class="reference external image-reference" href="http://www.gnu.org/licenses/agpl-3.0-standalone.html"><img alt="License: AGPL-3" src="https://img.shields.io/badge/licence-AGPL--3-blue.png" /></a> <a class="reference external image-reference" href="https://github.com/OCA/stock-logistics-shopfloor/tree/18.0/shopfloor"><img alt="OCA/stock-logistics-shopfloor" src="https://img.shields.io/badge/github-OCA%2Fstock--logistics--shopfloor-lightgray.png?logo=github" /></a> <a class="reference external image-reference" href="https://translation.odoo-community.org/projects/stock-logistics-shopfloor-18-0/stock-logistics-shopfloor-18-0-shopfloor"><img alt="Translate me on Weblate" src="https://img.shields.io/badge/weblate-Translate%20me-F47D42.png" /></a> <a class="reference external image-reference" href="https://runboat.odoo-community.org/builds?repo=OCA/stock-logistics-shopfloor&amp;target_branch=18.0"><img alt="Try me on Runboat" src="https://img.shields.io/badge/runboat-Try%20me-875A7B.png" /></a></p>
```
jbaudoux: Drop changes to auto generated files on imp/fix commit

shopfloor/actions/packing.py
henrybackman: Makes sense, I've changed to filtered_domain :+1:

shopfloor/services/checkout.py
henrybackman: Refactored to not need to pass the carrier and get it inside the packing action when needed with the packing action's _get_carrier method

shopfloor/services/checkout.py
henrybackman: Done

shopfloor/actions/packing.py
henrybackman: I added clarification and also noticed that should convert the "base_on_rule" delivery_type as well. In `stock_delivery` module these are considered as not real carrier integrations and are therefore converted to "none" when using the `choose.delivery.package` wizard
[https://github.com/odoo/odoo/blob/a7aa7a8aab975e53fc843ffdc5688e5672c3833c/addons/stock_delivery/models/stock_picking.py#L146](https://github.com/odoo/odoo/blob/a7aa7a8aab975e53fc843ffdc5688e5672c3833c/addons/stock_delivery/models/stock_picking.py#L146)

shopfloor/README.rst
henrybackman: The pre-commit added those :man_shrugging:

shopfloor/__manifest__.py
henrybackman: Yes, for the `package_type_domain` field. I used similar approach as the `shopfloor_cluster_picking_repack` `_get_available_delivery_package_type` to use the wizards domain to filter the package types
[https://github.com/OCA/stock-logistics-shopfloor/blob/1913aa5cda7345cce7fe2768138e06b18b63fed9/shopfloor_cluster_picking_repack/services/cluster_picking.py#L249](https://github.com/OCA/stock-logistics-shopfloor/blob/1913aa5cda7345cce7fe2768138e06b18b63fed9/shopfloor_cluster_picking_repack/services/cluster_picking.py#L249)

shopfloor/static/description/index.html
henrybackman: Added these to separate commit :+1:

jbaudoux: Drop that commit. Your PR shouldn't be red due to this. The ocabot already manages it automatically.

jbaudoux (approved):
Thanks. Looks good

jbaudoux:
@mmequignon @TDu Good for you?

jbaudoux:
@nicolas-delbovier-acsone FYI

shopfloor/actions/packing.py
```diff
@@ -11,16 +11,41 @@ class PackingAction(Component):
...
-        return package_type.package_carrier_type in (
-            "none",
-            carrier.delivery_type,
-        )
+    def _delivery_type_for_carrier(self, carrier):
+        # "fixed" and "base_on_rule" are pricing rules, not real carrier integrations,
+        # so they have no package type restrictions, treat them as no carrier.
+        # See: odoo/src/addons/stock_delivery/models/stock_picking.py line 146-147
+        if not carrier or carrier.delivery_type in ("fixed", "base_on_rule", False):
+            return False
```
jbaudoux: When there is no carrier, the package type carrier type must be False, but when there is a carrier but no integration, the package type carrier type must be "none".
Also add "pricelist" for compatibility with `delivery_carrier_pricelist` module that we use.
The selection can't be False, it's a required field.

```suggestion
        if not carrier:
            return False
        elif carrier.delivery_type in ("fixed", "base_on_rule", "pricelist"):
            return "none"
```

jbaudoux (changes requested):

shopfloor/services/zone_picking.py
```diff
@@ -1304,7 +1304,7 @@ def _handle_pick_pack_same_time_for_package(self, move_line, package):
         carrier = picking.ship_carrier_id or picking.carrier_id
         if carrier:
             actions = self._actions_for("packing")
```
divad1196, line 1306: Why not use `_package_type_good_for_carrier` here?

Also, the following 2 functions have the same body style:
- `_package_type_good_for_carrier`
- `_get_available_package_type` 

2 lines: get action and call utility. Is that overkill ?

divad1196 (approved):
Just the change mentioned by @jbaudoux 
Theh rest looks good

shopfloor/actions/packing.py
henrybackman: I changed this to return the correct value for those delivery types

jbaudoux (approved):

shopfloor/actions/packing.py
```diff
@@ -11,16 +11,43 @@ class PackingAction(Component):
...
+        if not carrier:
+            return False
+        elif carrier.delivery_type in ("fixed", "base_on_rule", "pricelist"):
+            return "none"
+        return carrier.delivery_type
+
+    def _get_carrier(self, picking):
+        return picking.ship_carrier_id or picking.carrier_id
+
+    def _get_package_type_domain(self, picking):
```
jbaudoux, line 27: Can you adapt `_get_available_delivery_package_type` in `shopfloor_cluster_picking_repack` with this new method?

jbaudoux:
@henrybackman You can also remove test-requirements.txt, dependency is now merged

shopfloor/actions/packing.py
henrybackman: Yes, adapted now to use the new methods

jbaudoux (approved):

shopfloor/services/checkout.py
```diff
@@ -1204,22 +1192,13 @@ def _scan_package_action_from_none(self, picking, selected_lines, record, **kw):
     def _get_carrier(self, picking):
         return picking.ship_carrier_id or picking.carrier_id
 
-    def _package_type_good_for_carrier(self, package_type, carrier):
+    def _package_type_good_for_carrier(self, picking, package_type):
```
mmequignon:
```suggestion
    def _package_type_valid_for_carrier(self, picking, package_type):
```
Not blocking. Just to be consistent regarding naming.

shopfloor/services/checkout.py
```diff
@@ -1204,22 +1192,13 @@ def _scan_package_action_from_none(self, picking, selected_lines, record, **kw):
...
         actions = self._actions_for("packing")
-        return actions.package_type_valid_for_carrier(package_type, carrier)
+        return actions.package_type_valid_for_carrier(package_type, picking=picking)
```
mmequignon:
```suggestion
        return actions.package_type_valid_for_carrier(package_type, picking)
```
Picking isn't a named argument in that method.

shopfloor/services/checkout.py
```diff
@@ -1204,22 +1192,13 @@ def _scan_package_action_from_none(self, picking, selected_lines, record, **kw):
...
-            return model.search(
-                [("package_carrier_type", "=", False)],
-                order="name",
-            )
-        return model.search(
-            [("package_carrier_type", "=", carrier.delivery_type or "none")],
-            order="name",
-        )
+        actions = self._actions_for("packing")
+        return actions.available_package_types_for_picking(picking, order="name")
```
mmequignon:
```suggestion
        return actions.available_package_types_for_picking(picking)
```
`order="name"` is already the default.

shopfloor/actions/packing.py
```diff
@@ -11,16 +11,43 @@ class PackingAction(Component):
...
+        carrier = self._get_carrier(picking)
+        wizard_obj = self.env["choose.delivery.package"]
+        delivery_type = self._delivery_type_for_carrier(carrier)
+        wizard = wizard_obj.with_context(
+            current_package_carrier_type=delivery_type
+        ).new({"picking_id": picking.id})
+        return wizard.package_type_domain
+
+    def package_type_valid_for_carrier(self, package_type, picking):
```
mmequignon, line 36:
```suggestion
    def package_type_valid_for_transfer(self, package_type, picking):
```
This isn't a carrier.

shopfloor/services/zone_picking.py
mmequignon: That's a helper method. Definitely not overkill. I would add that method.

shopfloor/tests/test_checkout_scan_package_action.py
```diff
@@ -475,3 +469,72 @@ def test_put_in_pack(self):
...
+        packing_action = type(self.service._actions_for("packing"))
+        with mock.patch.object(
+            packing_action,
+            "_get_package_type_domain",
+            side_effect=[
+                [("id", "=", other_packaging.id)],
+                [("id", "=", packaging.id)],
+                [("id", "=", packaging.id)],
+            ],
+        ):
```
mmequignon: You're not testing the code if you're mocking it.
Just spend a few minutes to configure carriers, and test it for real please.

mmequignon (changes requested):
Tests aren't enough.
You're modifying an action, and not testing it properly.
Please add tests in `tests/test_actions_packing.py`

shopfloor_cluster_picking_repack/services/cluster_picking.py
```diff
@@ -258,22 +258,12 @@ def _get_available_delivery_package_type(self, picking) -> PackageType:
         The returned package types are ordered by number of parcels then
         by name.
         """
-        model = self.env["stock.package.type"]
         carrier = picking.ship_carrier_id or picking.carrier_id
```
jbaudoux: Use your _get_carrier

shopfloor/actions/packing.py
henrybackman: This is following the existing naming convention. Only changing this function would make it feel out of place to me and I'm not sure about trying to change the term carrier -> transfer globally as part of this PR.

shopfloor/tests/test_checkout_scan_package_action.py
henrybackman: Good point, I've reworked the tests to create real packagings and check the message from the action

PR #107, opened 2026-05-18
[18.0][IMP] shopfloor: confirmed zero check creates draft inventory

mmequignon:
Zero checks are triggered when odoo thinks stock for a location is 0 after an operation.
When user clicks on `location not empty`, it creates a draft inventory.
When user clicks on `location empty`, nothing happens, as we assume there's nothing to do when counted qty is right.

This PR creates an inventory with `inventory_quantity` with qty = 0 or 1 depending on whether user clicked on `location empty` or `location not empty`.

### DIFF
--- a/shopfloor/actions/inventory.py
+++ b/shopfloor/actions/inventory.py
@@ -22,9 +22,11 @@ class InventoryAction(Component):
         # see comment in models/stock_inventory.py
         return self.env["stock.quant"].with_context(_sf_inventory=True)
 
-    def create_draft_check_empty(self, location, product, ref=None, lot=None):
+    def create_draft_check_empty(
+        self, location, product, ref=None, lot=None, package=None
+    ):
         """Create a draft inventory for a product with a zero quantity"""
-        return self._create_draft_inventory(location, product, lot=lot)
+        return self._create_draft_inventory(location, product, package=package, lot=lot)
 
     def _inventory_exists(self, location, product, package=None, lot=None):
         """Return if an inventory for location and product exist"""
@@ -39,8 +41,12 @@ class InventoryAction(Component):
             domain.append(("lot_id", "=", lot.id))
         return self.inventory_model.search_count(domain)
 
-    def _get_existing_quant(self, location, product, package=None, lot=None, limit=1):
-        domain = [("location_id", "=", location.id), ("product_id", "=", product.id)]
+    def _get_existing_quant(
+        self, location, product=None, package=None, lot=None, limit=1
+    ):
+        domain = [("location_id", "=", location.id)]
+        if product is not None:
+            domain.append(("product_id", "=", product.id))
         if package is not None:
             domain.append(("package_id", "=", package.id))
         else:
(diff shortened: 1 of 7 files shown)


vvrossem (approved):

shopfloor/actions/inventory.py
```diff
@@ -63,7 +63,7 @@ def _create_draft_inventory(self, location, product, package=None, lot=None):
                     {
                         # Set a user to prevent the zero quant cleanup
                         "user_id": self.env.user.id,
-                        "inventory_quantity": 0,
+                        "inventory_quantity": qty,
```
jbaudoux: Never set a quantity on a draft inventory. That was removed as misleading in this commit https://github.com/OCA/stock-logistics-shopfloor/commit/5918a880bb96350044d429db34acebd6a5f1d1d6

shopfloor/actions/inventory.py
```diff
@@ -98,6 +98,12 @@ def create_control_stock(
         if not self._inventory_exists(location, product, package=package, lot=lot):
             self._create_draft_inventory(location, product, package=package, lot=lot)
 
+    def confirm_not_empty(self, location, product, ref=None, lot=None):
+        return self.create_draft_check_empty(location, product, qty=1, ref=ref, lot=lot)
+
+    def confirm_empty(self, location, product, ref=None, lot=None):
+        return self.create_draft_check_empty(location, product, qty=0, ref=ref, lot=lot)
```
jbaudoux: Do not create a draft inventory (= a request to count) when the user confirms the quantity is correct.

Check on the location that there are no draft inventory by checking there is no quant with a user_id. Check there is no quant with a quantity.
If this is the case, then you can create a draft inventory but afterwards you need to call `action_apply_inventory()` on the result quants to validate it.

jbaudoux (changes requested):

TDu:
@jbaudoux I updated the PR following your comments, does it look good for you now ?

shopfloor/actions/inventory.py
```diff
@@ -98,6 +98,17 @@ def create_control_stock(
         if not self._inventory_exists(location, product, package=package, lot=lot):
             self._create_draft_inventory(location, product, package=package, lot=lot)
 
+    def confirm_not_empty(self, location, product, ref=None, lot=None):
+        return self.create_draft_check_empty(location, product, ref=ref, lot=lot)
+
+    def confirm_empty(self, location, product, ref=None, lot=None):
+        quants = self._get_existing_quant(location, product, lot=lot, limit=None)
```
jbaudoux: It's the location that is empty, not the product or lot
```suggestion
        quants = self._get_existing_quant(location, limit=None)
```

shopfloor/actions/inventory.py
```diff
@@ -98,6 +98,17 @@ def create_control_stock(
...
+        if not quants.filtered(lambda quant: quant.user_id or quant.quantity):
+            quants = self.create_draft_check_empty(location, product, ref=ref, lot=lot)
+            quants.sudo().action_apply_inventory()
+            return quants
+        return self.create_draft_check_empty(location, product, ref=ref, lot=lot)
```
jbaudoux: Could be there are pos and neg quants that are not merged yet. So I remove my suggestion to check on the quantity.
```suggestion
        if quants.filtered(lambda quant: quant.user_id and quant.user_id != self.env.user)
            # the location is empty but there is still a planned check empty inventory
            # do nothing, someone will have to check
            return
        if not quants:
            quants = self.create_draft_inventory(location, product, package=package, lot=lot)
        else:
            quants.write(
                    {
                        "user_id": self.env.user.id,
                        "inventory_quantity": 0,
                        "inventory_quantity_set": True,
                        "inventory_date": fields.Date.today(),
                    }
            )
        quants.sudo().action_apply_inventory()
```

shopfloor/actions/inventory.py
```diff
@@ -98,6 +98,17 @@ def create_control_stock(
         if not self._inventory_exists(location, product, package=package, lot=lot):
             self._create_draft_inventory(location, product, package=package, lot=lot)
 
+    def confirm_not_empty(self, location, product, ref=None, lot=None):
+        return self.create_draft_check_empty(location, product, ref=ref, lot=lot)
+
+    def confirm_empty(self, location, product, ref=None, lot=None):
```
jbaudoux: package arg is missing on both

jbaudoux (changes requested):

shopfloor/actions/inventory.py
TDu: Indeed :+1:

shopfloor/actions/inventory.py
TDu: Updated

shopfloor/actions/inventory.py
TDu: Added

jbaudoux (approved):
Thanks @TDu . Looks good now :)

TDu:
Squashed

PR #109, opened 2026-05-22
[18.0][FIX] shopfloor_reception: fix scan lot domain

TDu:
Fix for this stack trace
```
...
  File "/odoo/src/odoo/osv/expression.py", line 1459, in parse
    push_result(model._condition_to_sql(alias, left, operator, right, self.query))
                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/odoo/src/odoo/models.py", line 3150, in _condition_to_sql
    assert not isinstance(value, BaseModel), \
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
AssertionError: Invalid value stock.lot(126774,) in domain term ('name', '=', stock.lot(126774,))
```

### DIFF
--- a/shopfloor_reception/services/reception.py
+++ b/shopfloor_reception/services/reception.py
@@ -125,8 +125,8 @@ class Reception(Component):
             ("move_id.picking_id.state", "=", "assigned"),
             ("move_id.picking_id.user_id", "=", False),
             "|",
-            ("lot_id.name", "=", lot),
-            ("lot_name", "=", lot),
+            ("lot_id", "=", lot.id),
+            ("lot_name", "=", lot.name),
         ]
 
     def _domain_stock_picking(self, today_only=False):
--- a/shopfloor_reception/tests/test_select_document.py
+++ b/shopfloor_reception/tests/test_select_document.py
@@ -178,3 +178,35 @@ class TestSelectDocument(CommonCase):
             data={"pickings": self._data_for_pickings(picking)},
             message=message,
         )
+
+    def test_scan_lot_1(self):
+        picking = self._create_picking()
+        lot = self._create_lot()
+        line_product_a = picking.move_ids.move_line_ids.filtered(
+            lambda line: line.product_id == self.product_a
+        )
+        line_product_a.lot_id = lot
+        response = self.service.dispatch("scan_document", params={"barcode": lot.name})
+        body = "Several transfers found, please select a transfer manually."
+        self.assert_response(
+            response,
+            next_state="select_document",
+            data={"pickings": self._data_for_pickings(picking)},
+            message={"message_type": "error", "body": body},
+        )
+
+    def test_scan_lot_2(self):
+        picking = self._create_picking()
+        lot = self._create_lot()
+        line_product_a = picking.move_ids.move_line_ids.filtered(
+            lambda line: line.product_id == self.product_a
+        )
+        line_product_a.lot_name = lot.name
+        response = self.service.dispatch("scan_document", params={"barcode": lot.name})
+        body = "Several transfers found, please select a transfer manually."
+        self.assert_response(
+            response,
+            next_state="select_document",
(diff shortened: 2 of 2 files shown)


shopfloor_reception/services/reception.py
```diff
@@ -120,13 +120,14 @@ def _domain_move_line_by_product(self, product):
         ]
 
     def _domain_move_line_by_lot(self, lot):
+        lot_name = lot.name if lot else False
```
jbaudoux: How can you ensure this method if you have no lot?

shopfloor_reception/tests/test_select_document.py
```diff
@@ -178,3 +178,16 @@ def test_scan_packaging_no_picking(self):
             data={"pickings": self._data_for_pickings(picking)},
             message=message,
         )
+
+    def test_scan_lot(self):
+        picking = self._create_picking()
+        lot = self._create_lot()
```
jbaudoux, line 184: How is this testing your change?

jbaudoux (changes requested):

shopfloor_reception/services/reception.py
TDu: This is just to be on the safe side.
The error that is being fixed is the domain creation. I have added the stack trace that led to this change

shopfloor_reception/tests/test_select_document.py
TDu: It is testing the domain being changed

shopfloor_reception/services/reception.py
mmequignon: I think this is `search._find_record_by_type` responsibility to find the right record.
This is used in scan document_only, if no lot is found, next type is tested against barcode.
However, this could be used somewhere else, and it's nice to double check that.
However, I don't quite see the point to search lines without lot when no lot is provided.
Not sure.

mmequignon (approved):

shopfloor_reception/services/reception.py
TDu: I understand your point, I will check.

TDu: I removed the safety belt.
@jbaudoux  Ok for you now ?

mmequignon: You could add a `lot.ensure_one()` to state that this method is meant to be used with a lot, period.

TDu: I don't think it is needed. The important part of this PR is fixing the wrong domain, that was not even included in a unit test.

shopfloor_reception/services/reception.py
```diff
@@ -125,8 +125,8 @@ def _domain_move_line_by_lot(self, lot):
             ("move_id.picking_id.state", "=", "assigned"),
             ("move_id.picking_id.user_id", "=", False),
             "|",
-            ("lot_id.name", "=", lot),
-            ("lot_name", "=", lot),
+            ("lot_id.name", "=", lot.name),
```
jbaudoux:
```suggestion
            ("lot_id", "=", lot),
```

jbaudoux (changes requested):

jbaudoux:
cc @nicolas-delbovier-acsone

shopfloor_reception/tests/test_select_document.py
```diff
@@ -178,3 +178,16 @@ def test_scan_packaging_no_picking(self):
...
+        picking.move_line_ids[0].lot_id = lot
```
jbaudoux: Can you test also the lot_name case?

TDu: done

jbaudoux (approved):
Great

Can you squash the fixup?

PR #111, opened 2026-06-08
[18.0][FIX] shopfloor: fix cluster picking scan destination

TDu:
The `quantity` parameter for `scan_destination_pack` allows null values, but then calls `_split_qty_to_be_done` without intializing the quantity value.
And the endpoint crashes.

This changes makes the quantity parameter required for the endpoint, to see if it is a viable solution.

### DIFF
--- a/shopfloor/services/cluster_picking.py
+++ b/shopfloor/services/cluster_picking.py
@@ -1464,7 +1464,6 @@ class ShopfloorClusterPickingValidator(Component):
             "quantity": {
                 "coerce": to_float,
                 "required": True,
-                "nullable": True,
                 "type": "float",
             },
         }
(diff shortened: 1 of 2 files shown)


jbaudoux (commented):
Makes sense.

Have you tested with the option "no prefill" what happen if you don't scan the product to leave the quantity initialized to "0" and directly scan the destination pack?
I think you added the nullable in 2022 for that purpose. 

See https://github.com/OCA/wms/pull/483/changes#diff-816116d5e60c166e5b61ca80edf3948294ac329425f2c921d8af95fe5cc03167R1276

TDu:
> Makes sense.
> 
> Have you tested with the option "no prefill" what happen if you don't scan the product to leave the quantity initialized to "0" and directly scan the destination pack? I think you added the nullable in 2022 for that purpose.
> 
> See https://github.com/OCA/wms/pull/483/changes#diff-816116d5e60c166e5b61ca80edf3948294ac329425f2c921d8af95fe5cc03167R1276

I tried to reproduce this case, but I don't understand how I could scan the destination pack without scanning something before (to increase the quantity)

PR #112, opened 2026-06-12
[18.0][IMP] shopfloor_single_product_transfer: Improve lot scan uniqueness o…

lmignon:
…n location

Before this commit, when scanning a lot for a single product transfer, the system searched for the lot name without restricting it to the products available in the scanned location. This could lead to issues when there were multiple lots with the same name in different locations.

With this commit, the search for the lot name is now restricted to the products available in the scanned location. This ensures that the correct lot is identified during the scanning process, even if there are duplicate lot names in different locations. (At least if products in the location doesn't share lot names, but at least it reduces the risk of picking the wrong lot)

### DIFF
--- a/shopfloor_single_product_transfer/services/single_product_transfer.py
+++ b/shopfloor_single_product_transfer/services/single_product_transfer.py
@@ -845,16 +845,29 @@ class ShopfloorSingleProductTransfer(Component):
         package = self.env["stock.quant.package"].browse(package_id)
         if not location.exists() and not package.exists():
             return self._response_for_select_location_or_package()
+        products = (
+            self.env["stock.quant"]
+            .search([("location_id", "=", location.id or package.location_id.id)])
+            .product_id
+        )
         handlers_by_type = {
             "product": self._scan_product__scan_product,
             "packaging": self._scan_product__scan_packaging,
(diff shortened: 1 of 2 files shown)


shopfloor_single_product_transfer/services/single_product_transfer.py
```diff
@@ -845,16 +845,29 @@ def scan_product(self, barcode, location_id=None, package_id=None):
         package = self.env["stock.quant.package"].browse(package_id)
         if not location.exists() and not package.exists():
             return self._response_for_select_location_or_package()
+        products = (
+            self.env["stock.quant"]
+            .search([("location_id", "=", location_id)])
```
jbaudoux: The method accepts as argument a location_id or package_id
```suggestion
            .search([("location_id", "=", location.id or package.location_id.id)])
```

jbaudoux (changes requested):

shopfloor_single_product_transfer/services/single_product_transfer.py
lmignon: Good catch. Fixed

jbaudoux (approved):

TDu (approved):

PR #113, opened 2026-06-12
[IMP] shopfloor: remove dependency on stock_storage_type

jbaudoux:
Create new test data instead of relying on stock_storage_type
Drop dependency stock_storage_type

cc @TDu @mmequignon @simahawk @rousseldenis @lmignon

### DIFF
--- a/shopfloor/services/single_pack_transfer.py
+++ b/shopfloor/services/single_pack_transfer.py
@@ -201,11 +201,6 @@ class SinglePackTransfer(Component):
         )
         picking.action_confirm()
         picking.action_assign()
-        # For packages that contain several products (so linked to several
-        # moves), the putaway destination computation of the strategy
-        # triggered by `action_assign()` above won't work, so we trigger
-        # the computation manually here at the package level.
-        package_level.recompute_pack_putaway()
         return package_level
 
     def _is_move_state_valid(self, moves):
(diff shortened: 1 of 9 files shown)


shopfloor/services/single_pack_transfer.py
```diff
@@ -201,11 +201,6 @@ def _create_package_level(self, package):
         )
         picking.action_confirm()
         picking.action_assign()
-        # For packages that contain several products (so linked to several
-        # moves), the putaway destination computation of the strategy
-        # triggered by `action_assign()` above won't work, so we trigger
-        # the computation manually here at the package level.
-        package_level.recompute_pack_putaway()
```
jbaudoux, line 208: Removed because this should not be necessary anymore since odoo 15.0

shopfloor/tests/test_checkout_change_package_type.py
```diff
@@ -8,7 +8,7 @@ class CheckoutListSetPackageTypeCase(CheckoutCommonCase):
     @classmethod
     def setUpClassBaseData(cls):
         super().setUpClassBaseData()
-        cls.env["stock.package.type"].sudo().search([]).active = False
+        cls.env["stock.package.type"].sudo().search([]).unlink()
```
jbaudoux, line 11: Because there is no `active` field in model in base stock module

lmignon (approved):

shopfloor/tests/test_checkout_change_package_type.py
jbaudoux: Failure should be fixed with https://github.com/OCA/stock-logistics-putaway/pull/30

TDu (approved):

PR #116, opened 2026-06-12
[18.0][IMP] shopfloor_single_product_transfer: Back to select_location_or_p…

lmignon:
…ackage

When the user sets a location for a move line, we used to go back to the select_product state. This is not ideal in the case where the no more product needs to be selected at the current location (no more pending move line at this location), as the user would have to select a new location to be able to select the next move line.

Now, if the scenario is configured to forbid the creation of new move lines and there is no more pending move line at the current location, when the user sets a location for a move line, the next state will be select_location_or_package instead of select_product.

### DIFF
--- a/shopfloor_single_product_transfer/services/single_product_transfer.py
+++ b/shopfloor_single_product_transfer/services/single_product_transfer.py
@@ -70,8 +70,10 @@ class ShopfloorSingleProductTransfer(Component):
 
     # Responses
 
-    def _response_for_select_location_or_package(self, message=None):
-        return self._response(next_state="select_location_or_package", message=message)
+    def _response_for_select_location_or_package(self, message=None, popup=None):
+        return self._response(
+            next_state="select_location_or_package", message=message, popup=popup
+        )
 
     def _response_for_select_product(
         self,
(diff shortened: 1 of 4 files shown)


shopfloor_single_product_transfer/services/single_product_transfer.py
```diff
@@ -663,6 +678,13 @@ def _set_quantity__post_move(self, move_line, location, confirmation=None):
         message = self.msg_store.transfer_done_success(move_line.picking_id)
         completion_info = self._actions_for("completion.info")
         completion_info_popup = completion_info.popup(move_line)
+        if (
+            not self.is_allow_move_create()
+            and not self._has_pending_operations_at_same_location(move_line)
```
jbaudoux, line 683: Could you pass `location` to the method ? That would allow to have a move_line with a reservation on a sublocation of `location`

lmignon: but location is the destination location 🤔

jbaudoux: oups...

jbaudoux (approved):

TDu (approved):

PR #117, opened 2026-06-15
[18.0][IMP] shopfloor_single_product_transfer: Implement allow_get_work

lmignon:
If the menu allow it, at scenario start, the user will be able to request work to do in place of starting by scanning a location or a package.

This is useful in scenarios where the user might not know beforehand what they need to scan, but want to get an assignment first.

The UI part is done in #118

### DIFF
--- a/shopfloor_single_product_transfer/services/single_product_transfer.py
+++ b/shopfloor_single_product_transfer/services/single_product_transfer.py
@@ -82,6 +82,18 @@ class ShopfloorSingleProductTransfer(Component):
             )
         return self._response_for_select_location_or_package(message=message)
 
+    def _response_for_start_line(self, move_line, message=None):
+        """Transition to the 'start_line' state
+
+        This is used to confirm the processing of a move line
+        by the user. The user will be requested to select the
+        product or the package to process the move line.
+        """
+        data = {
+            "move_line": self.data.move_line(move_line),
+        }
+        return self._response(next_state="start_line", data=data, message=message)
+
     def _response_for_select_location_or_package(self, message=None, popup=None):
         return self._response(
             next_state="select_location_or_package", message=message, popup=popup
@@ -847,6 +859,38 @@ class ShopfloorSingleProductTransfer(Component):
             response = self._response_for_set_quantity(move_line, message=message)
         return response
 
+    def _scan_line__by_package(self, package, move_line):
+        if move_line.package_id == package:
+            return self._response_for_set_quantity(move_line)
+
+    def _scan_line__by_product(self, product, move_line):
+        if product == move_line.product_id:
+            if product.tracking in ("lot", "serial"):
+                return self._response_for_start_line(
+                    move_line,
+                    message=self.msg_store.scan_lot_on_product_tracked_by_lot(),
+                )
+            else:
+                return self._response_for_set_quantity(move_line)
+
+    def _scan_line__by_packaging(self, packaging, move_line):
+        return self._scan_line__by_product(packaging.product_id, move_line)
+
+    def _scan_line__by_lot(self, lot, move_line):
+        if lot == move_line.lot_id:
+            return self._response_for_set_quantity(move_line)
+
+    def _scan_line__fallback(self, record, move_line):
+        # Nothing matches what is expected from the move line.
+        if record:
+            return self._response_for_start_line(
+                move_line,
+                message=self.msg_store.wrong_record(record),
+            )
+        return self._response_for_start_line(
+            move_line, message=self.msg_store.barcode_not_found()
+        )
+
     # Endpoints
 
     def start(self):
@@ -858,12 +902,13 @@ class ShopfloorSingleProductTransfer(Component):
 
         First recover any started pickings.
         The find the first move line from the oldest transfer that can be worked on.
-        Mark all move lines on that location as picked.
+        Mark the first move lines as picked.
         And ask the user to confirm.
 
         Transitions:
         * start: no work found
-        * scan_location: with the location to work form for confirmation
+        * select_line: a move line has been found and marked as picked,
+            ask the user to confirm
         """
         response = self._recover_previous_session()
         if response:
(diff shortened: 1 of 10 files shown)


shopfloor_single_product_transfer/services/single_product_transfer.py
```diff
@@ -813,14 +825,51 @@ def _scan_location_or_package__by_location(self, location):
...
+    def start(self):
+        response = self._recover_previous_session()
+        return response or self._response_for_start()
+
+    def find_work(self):
+        """Find the new location to work from, for a user.
+
+        First recover any started pickings.
+        The find the first move line from the oldest transfer that can be worked on.
+        Mark all move lines on that location as picked.
```
jbaudoux:
```suggestion
        Mark the first move lines as picked.
```

shopfloor_single_product_transfer/services/single_product_transfer.py
```diff
@@ -813,14 +825,51 @@ def _scan_location_or_package__by_location(self, location):
...
+        * scan_location: with the location to work form for confirmation
+        """
+        response = self._recover_previous_session()
+        if response:
+            return response
+        self._actions_for("lock").advisory(self._advisory_lock_find_work)
+        move_lines = self.search_move_line.search_move_lines(match_user=True)
+        if not move_lines:
+            return self._response_for_start(message=self.msg_store.no_work_found())
+        location = fields.first(move_lines).location_id
```
jbaudoux:
```suggestion
        move_line = fields.first(move_lines)
        stock.mark_move_line_as_picked(move_line, quantity=0)
```

shopfloor_single_product_transfer/services/single_product_transfer.py
```diff
@@ -813,14 +825,51 @@ def _scan_location_or_package__by_location(self, location):
...
+        response = self._recover_previous_session()
+        return response or self._response_for_start()
+
+    def find_work(self):
+        """Find the new location to work from, for a user.
+
+        First recover any started pickings.
+        The find the first move line from the oldest transfer that can be worked on.
+        Mark all move lines on that location as picked.
+        And ask the user to confirm.
```
jbaudoux, line 1065: >         And ask the user to confirm.

It would be nice but you don't have such screen to confirm the selection. When you introduced get work in the location content transfer, you were not asking the user to confirm neither

shopfloor_single_product_transfer/services/single_product_transfer.py
```diff
@@ -813,14 +825,51 @@ def _scan_location_or_package__by_location(self, location):
...
+        """
+        response = self._recover_previous_session()
+        if response:
+            return response
+        self._actions_for("lock").advisory(self._advisory_lock_find_work)
+        move_lines = self.search_move_line.search_move_lines(match_user=True)
+        if not move_lines:
+            return self._response_for_start(message=self.msg_store.no_work_found())
+        location = fields.first(move_lines).location_id
+        return self._response_for_select_product(location=location)
```
jbaudoux: You cannot do that. If there are multiple products on the location, you can't let the user freely select any product/line.
You need here to return set quantity for the elected line

shopfloor_single_product_transfer/README.rst
```diff
@@ -1,7 +1,3 @@
-.. image:: https://odoo-community.org/readme-banner-image
```
jbaudoux, line 1: Drop changes to auto generated files

jbaudoux (changes requested):

shopfloor_single_product_transfer/services/single_product_transfer.py
jbaudoux: I don't think we can keep it without confirmation. A solution should be found.

lmignon:
@jbaudoux :sweat_smile: I wouldn't say it's perfect yet, but there have been many improvements. I've tried to take into account the various comments, including use cases that are quite different from my own.

shopfloor_single_product_transfer/services/single_product_transfer.py
```diff
@@ -218,7 +251,7 @@ def _scan_product__scan_product(self, product, location=None, package=None):
             self._scan_product__check_create_move_line,
             # First, try to create a move line with the available quantity
             self._scan_product__create_move_line,
-            # If no stock is available at first, try to unreserve moves if option
+            # If no stock is availfable at first, try to unreserve moves if option
```
jbaudoux: This change is probably not necessary 😄

shopfloor_single_product_transfer/services/single_product_transfer.py
```diff
@@ -813,14 +854,250 @@ def _scan_location_or_package__by_location(self, location):
...
+            selected_package_id=selected_package_id,
+        )
+
+    def _check_first_scan_location_or_pack_first(
+        self,
+        move_line,
+        selected_location_id=None,
+        selected_package_id=None,
+        scanned_location=None,
+        scanned_package=None,
```
jbaudoux, line 1001: Since I guess the scanned replaces the selected, why do you need here both arguments?
You could call it with the right argument

shopfloor_single_product_transfer/services/single_product_transfer.py
```diff
@@ -813,14 +854,250 @@ def _scan_location_or_package__by_location(self, location):
...
+    ):
+        """Restrict scanning product or lot first with option on.
+
+        When the option first scan location or pack first is on.
+        When the line being worked on has a package, asked to scan the package first.
+        When the line as a lot ask to scan the location first.
```
jbaudoux:
```suggestion
        When the line being worked on has a package, ask to scan the package first,
        otherwise ask to scan the location first.
```

shopfloor_single_product_transfer/services/single_product_transfer.py
```diff
@@ -813,14 +854,250 @@ def _scan_location_or_package__by_location(self, location):
...
+        if not self.work.menu.scan_location_or_pack_first:
+            return None
+        message = None
+        if move_line.package_id:
+            if not selected_package_id and not scanned_package:
+                message = self.msg_store.line_has_package_scan_package()
+        elif not selected_location_id and not scanned_location:
+            message = self.msg_store.scan_the_location_first()
+        if message:
+            return self._response_for_start_line(
```
jbaudoux, line 1018: If you do this in the caller instead of here, then this method would be generic and could move to shopfloor service.py
Just return message here

rousseldenis (approved):
LGTM

shopfloor_single_product_transfer/services/single_product_transfer.py
lmignon: both are required when a confirmation step is used since the value from step -1 must be provided back into the response

TDu (approved):

PR #123, opened 2026-08-04
[FIX] shopfloor: cluster_picking set_destination_all: skip batch sani…

lmignon:
…ty check

When set_destination_all splits the remaining lines into a new picking and re-attaches it to the batch, the batch may already contain a done picking (the one processed first). Assigning batch_id then triggers stock_picking_batch's sanity check, which raises because the batch mixes a done picking with a draft/assigned one.

Skip the sanity check for this specific assignment via a context key, since the new picking is a legitimate split of a batch already in progress, not an inconsistent manual assignment.

Forward port from https://github.com/OCA/wms/pull/1228

### DIFF
--- a/shopfloor/models/stock_move.py
+++ b/shopfloor/models/stock_move.py
@@ -137,10 +137,14 @@ class StockMove(models.Model):
                 if line.package_id == line.result_package_id:
                     line.result_package_id = False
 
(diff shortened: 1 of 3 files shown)


jbaudoux (approved):
/ocabot merge patch

PR #125, opened 2026-08-07
[18.0][IMP] shopfloor: add jump to menu functionality (zp, spt)

TDu:
Needs : 
* https://github.com/OCA/shopfloor-app/pull/28
* https://github.com/OCA/shopfloor-app/pull/29

### DIFF
--- a/shopfloor/models/shopfloor_menu.py
+++ b/shopfloor/models/shopfloor_menu.py
@@ -304,6 +304,14 @@ class ShopfloorMenu(models.Model):
         help="Python code to sort processing move lines.",
     )
 
+    jump_to_zone_picking_unload_all_menu_id = fields.Many2one(
+        comodel_name="shopfloor.menu", string="Jump after Unload All"
+    )
+    jump_to_single_pack_transfer_validate_menu_id = fields.Many2one(
+        comodel_name="shopfloor.menu",
+        string="Jump after Validate",
+    )
+
     @api.onchange("unload_package_at_destination")
     def _onchange_unload_package_at_destination(self):
         # Uncheck pick_pack_same_time when unload_package_at_destination is set to True
--- a/shopfloor/services/single_pack_transfer.py
+++ b/shopfloor/services/single_pack_transfer.py
@@ -62,6 +62,9 @@ class SinglePackTransfer(Component):
             message=message,
         )
 
+    def _get_data_for_jump_to_menu(self):
+        return self._response_for_start()
+
     def _scan_source(self, barcode, confirmation=None):
         """Search a package"""
         search = self._actions_for("search")
@@ -262,6 +265,9 @@ class SinglePackTransfer(Component):
         if self._is_last_move(move):
             completion_info = self._actions_for("completion.info")
             completion_info_popup = completion_info.popup(package_level.move_line_ids)
+        menu_jump = self.work.menu.jump_to_single_pack_transfer_validate_menu_id
+        if menu_jump:
+            return self._response_for_jump_to_menu(menu_jump, message=message)
         return self._response_for_start(message=message, popup=completion_info_popup)
 
     def _set_destination_and_done(self, package_level, scanned_location):
--- a/shopfloor/services/zone_picking.py
+++ b/shopfloor/services/zone_picking.py
@@ -303,6 +303,9 @@ class ZonePicking(Component):
             next_state="unload_set_destination", data=data, message=message
         )
 
+    def _get_data_for_jump_to_menu(self):
+        return self._response_for_start()
+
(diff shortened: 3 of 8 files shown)


shopfloor/services/single_pack_transfer.py
```diff
@@ -262,6 +262,11 @@ def _router_validate_success(self, package_level):
         if self._is_last_move(move):
             completion_info = self._actions_for("completion.info")
             completion_info_popup = completion_info.popup(package_level.move_line_ids)
+        menu_jump = self.work.menu._get_jumpto_menu(
+            "jump_to_menu_single_pack_transfer_validate"
+        )
+        if menu_jump:
```
simahawk, line 268:
```suggestion
        menu_jump = self.work.menu.jump_to_menu_single_pack_transfer_validate
        if menu_jump:
```

shopfloor/services/zone_picking.py
```diff
@@ -1267,7 +1267,14 @@ def set_destination(
...
                 message["body"] += "\n" + extra_message["body"]
 
-        # Process the next line
+        return self._process_next_line(message=message)
+
+    def _process_next_line(self, message=None):
+        menu_jump = self.work.menu._get_jumpto_menu(
+            "jump_to_menu_zone_picking_unload_all"
+        )
+        if menu_jump:
```
simahawk, line 1276:
```suggestion
        menu_jump = self.work.menu.jump_to_menu_zone_picking_unload_all
```

shopfloor/models/shopfloor_menu.py
```diff
@@ -304,6 +304,14 @@ class ShopfloorMenu(models.Model):
         help="Python code to sort processing move lines.",
     )
 
+    jump_to_menu_zone_picking_unload_all = fields.Many2one(
```
simahawk, line 307: we should suffix them w/ `_id` to respect the convention.

shopfloor/tests/test_single_pack_transfer_jump_to_menu.py
```diff
@@ -0,0 +1,106 @@
...
+
+    def assert_response_jump_to_menu(
+        self, response, menu, state_data=None, message=None
+    ):
+        data = {
+            "next_state": "start",
+            "menu_id": menu.id,
+        }
+        if state_data:
+            data["state_data"] = state_data
```
simahawk, line 88: should `states_data` w/ the inner mapping for the given state

simahawk: also the default jump_to_schema will force the serialization of the states data to JSON string

simahawk (changes requested):

shopfloor/services/single_pack_transfer.py
TDu: Changed

shopfloor/services/zone_picking.py
TDu: Changed

shopfloor/tests/test_single_pack_transfer_jump_to_menu.py
TDu: updated

simahawk (approved):

simahawk:
squash pls

shopfloor/services/zone_picking.py
```diff
@@ -303,6 +303,13 @@ def _response_for_unload_set_destination(
             next_state="unload_set_destination", data=data, message=message
         )
 
+    def _get_data_for_jump_to_menu(self):
+        response = self._response_for_start()
+        # The initial_state_key on the front end scenario is `scan_location`
```
simahawk: You don't need this. `start` is an alias for the initial state and it will be handled by the scenario mechanism.

It is similar to `init` but init is going to re-initialize everything including storage, messages, etc.

shopfloor/views/shopfloor_menu.xml
```diff
@@ -11,6 +11,26 @@
                     options="{'no_create': 1}"
                 />
             </field>
+            <group name="options" position="before">
```
simahawk: what about having a new tab in the base view instead?

simahawk (changes requested):

shopfloor/services/zone_picking.py
TDu: I have added this change after testing, because I hit this issue and that was my quick fix for it, I did not look deeper for the reason.
Before this change the user would land on `..../start` instead of `..../scan_location` as he does when accessing the menu direclty.

henrybackman (approved):

